PR #2, opened 2015-03-18
Feature/async tests

lorenzoPrimi:
@bassosimone can you check out to see how best to achieve what we are trying to do with @hellais. Basically we need to start the libevent loop in another thread since starting it in the current one leads to the UI being blocked (when calling libight_loop), but there seem to be some problems with doing so.

Perhaps this is something that we can add to libight directly.


bassosimone:
Hi @lorenzoPrimi!

I've poked around a bit with threads. The first thing I learned is that one should link not only to `libevent.a` but also to `libevent_pthreads.a` when thread safety is desired.

Also, one should call `evthread_use_pthreads()` (implemented in `libevent_pthreads.a`) so that libevent knows that pthreads must be used.

Apart from that, I fully agree with you that this functionality shall be implemented by libight rather than by each specific application and [I developed a working prototype of this](https://github.com/TheTorProject/libight/pull/104).

I've also looked at the code in this pull request but I am proceeding slow because I don't understand ObjC and iOS programming very well. If time permits, later this afternoon I'll try to run debug the code, otherwise I guess I cannot try to debug the code before Monday :(.

bassosimone:
Ok, closing this and reviewing #8.

PR #14, opened 2015-06-22
Run tests through async

bassosimone:
This pull request changes the way in which we run tests. Now we use the async interface exported by measurement-kit. It fixes the following issues:

* measurement-kit/measurement-kit-app-ios#4 ("Fix crash when running DNS injection")
* measurement-kit/measurement-kit-app-ios#9 ("Tests should be run using the async interface")

While there, I also simplified the search paths and tweaked compile settings to reduce warnings.

Please review. Once this is merged we can continue to work on other open issues such as how to correctly pass log messages to the graphical interfaces.


bassosimone:
I'm going to merge. Testing and merge okayed by @lorenzoPrimi.

PR #15, opened 2015-06-22
Allow multiple tests in a row (fix issues/7)

bassosimone:
This should fix measurement-kit/measurement-kit-app-ios#7 and allow one to submit two tests by clicking twice on RUN.


bassosimone:
Oops, this pull request has not the correct base, which should be master. Closing this, sorry.

PR #17, opened 2015-07-31
Autoreload table when test is over

bassosimone:
This pull request contains @lorenzoPrimi's patch to autoreload table when a test is over, plus another commit to sync up with measurement-kit-build-ios such that SDKVERSION="8.4" is used.


bassosimone:
Thanks :+1:!

PR #20, opened 2015-10-17
Cocoapod

AntonioLangiu:
I settled up the App to use cocoapod instead of build script. There are different improvement:

* no more subrepository to include measurement-kit
* no more build script, with the command **pod install** it download and build the frameworks. 
* to update the library you have just to remove the folders Pods and use the command **pod install**


AntonioLangiu:
@lorenzoPrimi can you please check it?

bassosimone:
Hi @AntonioLangiu,

Great work, thank you! I've tested it and this is what happened:

1. run `pod install` from the toplevel directory and this went smoothly

2. then I opened the `Libight_iOS.xcodeproj` project and built the project

This last step failed in the linking phase with the following error:

```
ld: library not found for -lPods
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

Any idea how to fix this? (The XCode project lists libPods.a as an unmet dependency - meaning that it's listed and the color is red.)

AntonioLangiu:
Hi @bassosimone,

Thank you. 

You have to install the cocoapod with `pod install` and then you have to open the `Libight_iOS.xcworkspace` instead of the `Libight_iOS.xcodeproj` that you opened. 

Is not an issue. I tested now the App with OS X El Capitan and Xcode 7.1.1 and it works!

bassosimone:
Thanks for clarifying. Now it compiles. Thank you, great work! :+1:

PR #27, opened 2016-02-12
Teach Podfile to actually use CocoaPods

bassosimone:
Before, the Podfile was just downloading from GitHub. Now, I would like to have a Podfile where, if you follow default instructions, you actually get measurement_kit to work as a dependency.

Therefore, I followed the default instructions. That is, I removed the original Podfile, then run `pod init`, then edited the new Podfile.

Now, when running `pod install`, installation is successful, except that pod spews out the following warning for each target:

```
[!] CocoaPods did not set the base configuration of your project because your project already has a custom config set. In order for CocoaPods integration to work at all, please either set the base configurations of the target `Libight_iOS` to `Pods/Target Support Files/Pods-Libight_iOS/Pods-Libight_iOS.debug.xcconfig` or include the `Pods/Target Support Files/Pods-Libight_iOS/Pods-Libight_iOS.debug.xcconfig` in your build configuration.
```

And, unsurprisingly, I am not able to build the project now.

Committing in any case what I did, such that we can further work on it.

@lorenzoPrimi could you help me here?


lorenzoPrimi:
Fixed in 2248cc79408f91e2f961529dbaea6f85654da6de

PR #32, opened 2016-06-07
Mk 0.2

bassosimone:
This branch is not ready to be merged yet. It contains changes to the app caused by updating MK to 0.2.0-beta. I have already applied changes so the app compiles. I have also smoke tested the code using the emulator and, while at it, I've started making a list of things to do next:

- [x] make sure files are actually written on disk so they are also sent to OONI's collector
- [x] add geoip files as resources so that geolocation works
- [x] understand whether dns_injection requires a DNS server to be used for auxiliary operations


NetProbe/Model/NetworkMeasurement.mm
```diff
@@ -14,43 +14,41 @@
...
+            continue;
+        }
+        NSLog(@"found DNS resolver: %s", addr);
+        dns_server = addr;
+        break;
+    }
+    free(res);
+    return dns_server;
+}
+
```
bassosimone: This is a rewrite of the algorithm above. Please check I made no stupid mistakes here :-).

NetProbe/Model/NetworkMeasurement.mm
```diff
@@ -87,10 +85,10 @@ - (void) run {
     NSBundle *bundle = [NSBundle mainBundle];
     NSString *path = [bundle pathForResource:@"hosts" ofType:@"txt"];
     mk::ooni::DnsInjectionTest()
-        .set_backend("8.8.8.8:53")
+        .set_options("backend", "8.8.8.1:53")
         .set_input_file_path([path UTF8String])
-        .set_verbose()
-        .on_log([self](const char *s) {
+        .set_verbosity(MK_LOG_DEBUG2)
+        .on_log([self](uint32_t, const char *s) {
```
bassosimone: What you see above is what changed in the API. Below there would be more noise because I also reindented.

NetProbe/Model/NetworkMeasurement.mm
```diff
@@ -120,24 +118,25 @@ -(id) init {
...
-            [self.logLines addObject:current];
-        });
-    })
-    .run([self]() {
-        NSLog(@"http_invalid_request_line testEnded");
-        dispatch_async(dispatch_get_main_queue(), ^{
-            self.finished = TRUE;
-            [[NSNotificationCenter defaultCenter] postNotificationName:@"refreshTable" object:nil];
+        .set_options("backend", "http://213.138.109.232/")
+        .set_options("dns/nameserver", get_dns_server())
```
bassosimone: I've added an explicit dns nameserver for http-invalid-request-line and tcp-connect. Not sure whether one is also needed for dns-injection, but perhaps yes. Other options that would be needed are the ones for setting the report file path and the location of the two databases used to geolocate.

bassosimone:
Added 9abd4a9 attempting to repair Travis!

NetProbe/Model/NetworkMeasurement.mm
```diff
@@ -53,21 +54,69 @@ @implementation NetworkMeasurement
...
+    NSArray *paths = NSSearchPathForDirectoriesInDomains(NSDocumentDirectory, NSUserDomainMask, YES);
+    NSString *documentsDirectory = [paths objectAtIndex:0];
+    //double timestamp = mk::time_now();
+    NSString *fileName = [NSString stringWithFormat:@"%@/test-%@.%@", documentsDirectory, self.test_id, ext];
+    //NSLog(@"fileName %@", fileName);
+    return fileName;
+}
+
+-(void)writeOrAppend:(NSString*)string{
+    NSString *fileName = [self getFileName:@"txt"];
```
bassosimone: Also here I'd name the file with `.log` extension

NetProbe/Model/NetworkMeasurement.mm
```diff
@@ -53,21 +54,69 @@ @implementation NetworkMeasurement
...
+    if(![fileManager fileExistsAtPath:fileName])
+    {
+        [string writeToFile:fileName atomically:NO encoding:NSStringEncodingConversionAllowLossy error:nil];
+    }
+    else
+    {
+        NSFileHandle *myHandle = [NSFileHandle fileHandleForWritingAtPath:fileName];
+        [myHandle seekToEndOfFile];
+        [myHandle writeData:[[NSString stringWithFormat:@"\n%@",string] dataUsingEncoding:NSUTF8StringEncoding]];
+    }
```
bassosimone: Is it correct to say that here we open the log file and seek to its end every time we have a new log line? Wouldn't it be more natural to open the file once and then write each log line into it?

Would it make sense to add support for writing the log file directly inside MK? (This change would be trivial)

NetProbe/Model/NetworkMeasurement.mm
```diff
@@ -53,21 +54,69 @@ @implementation NetworkMeasurement
...
+    if (!self) {
+        return nil;
+    }
+    self.name = [coder decodeObjectForKey:@"Test_name"];
+    self.test_id = [coder decodeObjectForKey:@"Test_id"];
+    self.json_file = [coder decodeObjectForKey:@"Test_jsonfile"];
+    self.log_file = [coder decodeObjectForKey:@"Test_logfile"];
+    self.completed = [[coder decodeObjectForKey:@"test_completed"] boolValue];
+    return self;
+}
```
bassosimone: Asking out of ignorance, what do `encodeWithCoder` and `initWithCoder` do? What are they encoding?

NetProbe/Model/NetworkMeasurement.mm
```diff
@@ -81,31 +130,39 @@ -(id) init {
 }
 
 - (void) run {
+    self.test_id = [NSNumber numberWithDouble:[[NSDate date] timeIntervalSince1970]];
+    self.json_file = [NSString stringWithFormat:@"test-%@.json", self.test_id];
+    self.log_file = [NSString stringWithFormat:@"test-%@.txt", self.test_id];
```
bassosimone: I'd name the log file with `.log` extension

NetProbe/Utility/TestStorage.h
```diff
@@ -0,0 +1,28 @@
+//
+//  TestStorage.h
+//  NetProbe
+//
+//  Created by Lorenzo Primiterra on 11/06/16.
+//  Copyright © 2016 Simone Basso. All rights reserved.
+//
```
bassosimone: I'd put here the standard comment we put at top of MK files

NetProbe/Utility/TestStorage.mm
```diff
@@ -0,0 +1,120 @@
+//
+//  TestStorage.m
+//  NetProbe
+//
+//  Created by Lorenzo Primiterra on 11/06/16.
+//  Copyright © 2016 Simone Basso. All rights reserved.
+//
```
bassosimone: I'd put here the comment we usually put at top of MK files

Podfile.lock
```diff
@@ -0,0 +1,28 @@
+PODS:
```
bassosimone: Uhm, why do we need to commit `Podfile.lock`? Isn't it specific to the local machine?

NetProbe/Model/NetworkMeasurement.mm
lorenzoPrimi: Used to save this class on disk (NSUserDefaults)
https://blog.soff.es/archiving-objective-c-objects-with-nscoding

NetProbe/Utility/TestStorage.h
lorenzoPrimi: This is XCode auto generated. What is the standard comment?

Podfile.lock
lorenzoPrimi: "Whether or not you check in the Pods directory, the Podfile and Podfile.lock should always be kept under version control."
https://guides.cocoapods.org/using/using-cocoapods.html

NetProbe/Model/NetworkMeasurement.mm
bassosimone: Ah, cool! So, this is a general way to store and then reload `NSObject` from a `.plist` file? Awesome!

NetProbe/Utility/TestStorage.h
bassosimone: This one:

```
// Part of measurement-kit <https://measurement-kit.github.io/>.
// Measurement-kit is free software. See AUTHORS and LICENSE for more
// information on the copying conditions.
```

Podfile.lock
bassosimone: Aw, snap! I didn't know... Thanks for educating me!

NetProbe/Model/NetworkMeasurement.mm
```diff
@@ -14,62 +15,106 @@
...
 
 -(id) init {
     self = [super init];
-    self.logLines = [[NSMutableArray alloc] init];
-    self.finished = FALSE;
+    NSBundle *bundle = [NSBundle mainBundle];
+    geoip_asn = [bundle pathForResource:@"GeoIPASNum" ofType:@"dat"];
+    geoip_country = [bundle pathForResource:@"GeoIP" ofType:@"dat"];
+    ca_cert = [bundle pathForResource:@"cacert" ofType:@"pem"];
+    self.completed = FALSE;
```
bassosimone: @lorenzoPrimi I have another ignorance-driver question here: my understanding is that this code sets the `geoip_asn` field and then also the `completed` field... so, why in the latter case `self` is prefixed to the attribute name and in the former case is is not? Is this related to us using properties?

NetProbe/Model/NetworkMeasurement.mm
bassosimone: I discussed this with @lorenzoPrimi and we agreed that for now we will leave this code in place, since it works, but that it would be optimal for MK to save logs (see measurement-kit/measurement-kit#706).

NetProbe/Model/NetworkMeasurement.mm
```diff
@@ -83,30 +128,39 @@ -(id) init {
...
-        .on_log([self](const char *s) {
+    mk::ooni::DnsInjection()
+        .set_options("backend", "8.8.8.1:53")
+        .set_options("dns/nameserver", get_dns_server())
+        .set_options("net/ca_bundle_path", [ca_cert UTF8String])
+        .set_options("geoip_country_path", [geoip_country UTF8String])
+        .set_options("geoip_asn_path", [geoip_asn UTF8String])
+        .set_input_filepath([path UTF8String])
+        .set_output_filepath([[self getFileName:@"json"] UTF8String])
+        .set_verbosity(MK_LOG_DEBUG2)
```
bassosimone: Note to self: we should probably lower the log level before merging the pull request because this level of verbosity is frankly very high (I think `MK_LOG_INFO` would be much better).

NetProbe/Model/NetworkMeasurement.mm
```diff
@@ -118,26 +172,36 @@ -(id) init {
...
+    mk::ooni::HttpInvalidRequestLine()
+        .set_options("backend", "http://213.138.109.232/")
+        .set_options("dns/nameserver", get_dns_server())
+        .set_options("net/ca_bundle_path", [ca_cert UTF8String])
+        .set_options("geoip_country_path", [geoip_country UTF8String])
+        .set_options("geoip_asn_path", [geoip_asn UTF8String])
+        .set_output_filepath([[self getFileName:@"json"] UTF8String])
+        .set_verbosity(MK_LOG_DEBUG2)
+        .on_log([self](uint32_t, const char *s) {
+            // XXX OK to send messages to object from another thread?
```
bassosimone: Note to self: we can remove this comment because we use `dispatch_async`

NetProbe/Model/NetworkMeasurement.mm
```diff
@@ -151,30 +215,38 @@ -(id) init {
...
-            [self.logLines addObject:current];
-        });
-    })
-    .run([self]() {
-        NSLog(@"tcp_connect testEnded");
-        dispatch_async(dispatch_get_main_queue(), ^{
-            self.finished = TRUE;
-            [[NSNotificationCenter defaultCenter] postNotificationName:@"refreshTable" object:nil];
+    mk::ooni::TcpConnect()
+        .set_options("port", "80")
```
bassosimone: Note to self: here we can pass port as integer

NetProbe/Utility/TestStorage.h
bassosimone: I will replace the header as part of merging this pull request.

NetProbe/Model/NetworkMeasurement.mm
bassosimone: As part of merging this pull request, I will create a issue reminding us of that.

NetProbe/Utility/TestStorage.mm
bassosimone: I'll fix this while merging

NetProbe/Utility/TestStorage.mm
```diff
@@ -0,0 +1,120 @@
...
+
++ (NSArray*)remove_test:(NSNumber*)test_id{
+    [self checkTests];
+    NSMutableArray *cache = [[[NSUserDefaults standardUserDefaults] arrayForKey:@"tests"] mutableCopy];
+    for (int i = 0; i < [cache count]; i++) {
+        NetworkMeasurement* test = [NSKeyedUnarchiver unarchiveObjectWithData:[cache objectAtIndex:i]];
+        if (test.test_id == test_id){
+            [self removeFile:test.json_file];
+            [self removeFile:test.log_file];
+            [cache removeObjectAtIndex:i];
```
bassosimone: Here I think you should `break` since you are looking for just one object inside the cache, or am I missing something obvious? If instead you want to continue iterating over the array, wouldn't you skip one element because you remove the element at `i`, then element `i+1` becomes `i` but you next iteration (if any) is on `i+1` (i.e. `i+2` before the element was removed?

NetProbe/Utility/TestStorage.mm
```diff
@@ -0,0 +1,120 @@
...
+//
+//  Created by Lorenzo Primiterra on 11/06/16.
+//  Copyright © 2016 Simone Basso. All rights reserved.
+//
+
+#import "TestStorage.h"
+
+@implementation TestStorage
+
++ (NSArray*)get_tests{
```
bassosimone: Shouldn't this function call `[self checkTests]` such that we always have an array to return? Or, is there a case where it makes sense to return `nil` to signal something to the caller?

NetProbe/Utility/TestStorage.mm
```diff
@@ -0,0 +1,120 @@
...
+    [[NSUserDefaults standardUserDefaults] setObject:cache forKey:@"tests"];
+    [[NSUserDefaults standardUserDefaults] synchronize];
+    return cache;
+}
+
++ (NSArray*)remove_test_atindex:(long)index{
+    [self checkTests];
+    NSMutableArray *cache = [[[NSUserDefaults standardUserDefaults] arrayForKey:@"tests"] mutableCopy];
+    if ([cache count] > index) {
+        NetworkMeasurement* test = [NSKeyedUnarchiver unarchiveObjectWithData:[cache objectAtIndex:index]];
```
bassosimone: I was wondering how `objectAtIndex` would react if passed a negative value

NetProbe/Utility/TestStorage.mm
```diff
@@ -0,0 +1,120 @@
...
+    }
+    return nil;
+}
+
++ (void)set_completed:(NSNumber*)test_id {
+    [self checkTests];
+    NSMutableArray *cache = [[[NSUserDefaults standardUserDefaults] arrayForKey:@"tests"] mutableCopy];
+    for (int i = 0; i < [cache count]; i++) {
+        NetworkMeasurement* test = [NSKeyedUnarchiver unarchiveObjectWithData:[cache objectAtIndex:i]];
+        if ([test.test_id isEqualToNumber:test_id]){
```
bassosimone: Here you use this construct to check for equality, above you use `test.test_id == test_id`... I am wondering (again this is ignorance!) if they are equivalent!

NetProbe/Utility/TestStorage.mm
```diff
@@ -0,0 +1,120 @@
...
+            test.completed = TRUE;
+            [cache setObject:[NSKeyedArchiver archivedDataWithRootObject:test] atIndexedSubscript:i];
```
bassosimone: Also here I think you would like to `break` as soon as the entry is found

NetProbe/View/LogViewController.mm
```diff
@@ -16,7 +16,32 @@ @implementation LogViewController
...
+        [self.logView setText:prettyJson];
+    }
+}
+
+-(NSString*)prettyJson:(NSString*)jsonString{
+    NSError *error;
+    NSData *jsonData = [jsonString dataUsingEncoding:NSUTF8StringEncoding];
+    id jsonObject = [NSJSONSerialization JSONObjectWithData:jsonData options:NSJSONReadingMutableContainers error:&error];
+    if (jsonObject){
+        NSData *prettyJsonData = [NSJSONSerialization dataWithJSONObject:jsonObject options:NSJSONWritingPrettyPrinted error:&error];
```
bassosimone: Here, shouldn't you check for errors?

bassosimone:
Okay, I did another reading of the diff and added more comments! Now I am going to see whether I could fix travis-ci and afterwards will proceed with testing on device!

bassosimone:
- [ ] fix travis
- [ ] apply small fixes
- [ ] on device tests

bassosimone:
Okay, I finished testing the App. 👍

I'll happily merge once my last outstanding comments are addressed.

NetProbe/Model/NetworkMeasurement.mm
lorenzoPrimi: check the .h file to understand why.
some var are exposed to the outside with @property

NetProbe/Model/NetworkMeasurement.mm
lorenzoPrimi: yes

NetProbe/Model/NetworkMeasurement.mm
lorenzoPrimi: Changed

NetProbe/Model/NetworkMeasurement.mm
lorenzoPrimi: done

NetProbe/Utility/TestStorage.h
lorenzoPrimi: ok so I will leave this to you

NetProbe/Utility/TestStorage.mm
lorenzoPrimi: yes, right

NetProbe/Utility/TestStorage.mm
lorenzoPrimi: I will add the return part here. check it when I push

NetProbe/Utility/TestStorage.mm
lorenzoPrimi: Can't happen as the index is the table row

NetProbe/Utility/TestStorage.mm
lorenzoPrimi: better  to use isEqualToNumber, I will change the other

NetProbe/Utility/TestStorage.mm
lorenzoPrimi: done, check it

NetProbe/Model/NetworkMeasurement.mm
```diff
@@ -14,62 +15,108 @@
 static void setup_idempotent() {
     static bool initialized = false;
     if (!initialized) {
-
         // Set the logger verbose and make sure it logs on the "logcat"
-        mk::set_verbose(1);
-        // XXX Ok to call NSLog() from another thread?
```
bassosimone: For the records: `NSLog()` is thread safe. Source: https://developer.apple.com/library/mac/documentation/Cocoa/Conceptual/Multithreading/ThreadSafetySummary/ThreadSafetySummary.html

NetProbe/Model/NetworkMeasurement.mm
bassosimone: ok, thanks!

bassosimone:
Tested again with MK v0.2.4! It works! 👍 

Now, merging!

PR #39, opened 2016-10-04
Informed consent

lorenzoPrimi:
Implemented informed consent


bassosimone:
So, I have tested this on a real device. Since when we started this pull request there was the migration to the new version of Xcode and this led to significant changes in my local project files. I'd rather avoid committing such changes because they are probably mechanical and so better to not include them here to reduce the signal / noise ratio. There will be time to commit those next.

My understanding is that this pull request is meant to be an initial implementation of the informed consent and that several things should be improved (e.g. the wording and other possibly not directly related with informed consent aspect that I have been addressed by @hellais in other issues).

As such, I bless this branch and will proceed with merging! Thank you!

(I am going to squash because it was not explicitly stated that the history shall be preserved and it seems to me the general situation is such that squash would be okay.)

PR #62, opened 2017-01-24
Update web ui

hellais:
@lorenzoPrimi you can use this as the new web UI.

I also updated the podfile to pull in the measurement-kit ndt-stats branch.


hellais:
New Web UI:
```
2.8M	webui
```

Old Web UI:
```
5.1M	webui
```

PR #77, opened 2017-02-06
Show context info along with progress

bassosimone:
The ideal progress bar should never stutter. Probably, a real progress bar sometimes does because some operations (e.g. connecting to a host) are atomic and they may stall.

I think that printing what the app is doing below/above the progress bar reduces stuttering-induced anxiety, especially if possibly long operations are marked as such, for example:

> submitting results; please be patient...

Additionally, printing what the app does leads to continuous user awareness. One may be aware after reading the intro and then forget. More difficult to forget, if you see messages like:

> testing http://controversial.com for censorship

I think this (i.e. provide a small looking glass to users) is a desirable property.

I am wondering whether doing that makes the design less neat (I do really like it!): iOS does not add text to the progress bar, but Ubuntu (that I think is also quite pretty) does.

install_ubuntu_10_04_from_usb_stick

This pull request implements what I'm proposing, in a ghetto way, to show it's possible:

simulator screen shot 6 feb 2017 15 21 08

Thanks to @lorenzoPrimi who helped me to understand what was the best and quickest way to implement this change :-). If there's consensus, we can work on making it pretty!

### DIFF
--- a/ooniprobe/Base.lproj/Main.storyboard
+++ b/ooniprobe/Base.lproj/Main.storyboard
@@ -128,20 +128,20 @@
         <scene sceneID="n9F-Ud-ZKt">
             <objects>
                 <tableViewController id="RZH-Z4-d2a" customClass="RunTestsViewController" sceneMemberID="viewController">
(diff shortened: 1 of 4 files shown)


bassosimone:
I've discussed this with @ebdet2002, and he made the very good point that, if all tests are running at once, the UX would be very messy. In general, he says, we should not provide textual info unless the majority of users asks explicitly asks us to do so. I guess this means *icebox*.

bassosimone:
Since there was no consensus and @ebdet2002 raised very valid points, I guess it makes sense to close this pull request at this point. We will implement a smoother progress bar.

PR #84, opened 2017-02-09
[readme] add link to App Store

ReadmeCritic:


ReadmeCritic:
ok

ReadmeCritic:
@mrphs updated

screen shot 2017-02-09 at 3 04 32 pm

hellais:
@ReadmeCritic thanks for suggesting this!

I am going to merge this and move the image from torproject.org to github so that we can avoid the potential blocking of torproject.org as @mrphs suggested.

PR #117, opened 2017-07-25
Add spec for mobile deep linking support

hellais:


bassosimone:
LGTM!

hellais:
> run_test has been renamed in nettests?
> ooniprobe:// has been renamed in ooni://
> Just asking because some work has already been done based on my specs and these are different. Can I change them?

I don't have a problem with changing `nettests` to `run_test`, nor with changing `ooni` to `ooniprobe`. The main reason for doing it like that was to have the URI be a bit more compact (though no conclusive evidence is brought on this, there is some fear and doubt being spread online: https://stackoverflow.com/questions/11546582/custom-url-scheme-maximum-url-length), so if we can save up some space, why not?

PR #119, opened 2017-09-07
Feature/uri scheme

lorenzoPrimi:


hellais:
Does this support Universal Link?

Do you have an `apple-app-site-association` file (https://developer.apple.com/library/content/documentation/General/Conceptual/AppSearch/UniversalLinks.html) for me to upload to `run.ooni.io`?

hellais:
@lorenzoPrimi in the future can you keep the changes related to code separate from the ones just updating translations (that should maybe only happen before the release is ready to be done)? It makes reading the diff a bit easier.

lorenzoPrimi:
No there isn't support for Universal links, I guessed you had to make some changes in the backend before.

lorenzoPrimi:
The translations were only updated at the end, yesterday.

hellais:
> No there isn't support for Universal links, I guessed you had to make some changes in the backend before.

To my knowledge the only thing I need to do on the "backend" is upload the `apple-app-site-association` file. Do I need to do something else?

Podfile.lock
```diff
@@ -14,31 +14,31 @@ DEPENDENCIES:
...
+  - measurement_kit (from `https://github.com/measurement-kit/measurement-kit.git`, tag `v0.7.1`)
   - PBRevealViewController
   - Toast
   - UIBarButtonItem-Badge-Coding
 
 EXTERNAL SOURCES:
   measurement_kit:
     :git: https://github.com/measurement-kit/measurement-kit.git
-    :tag: v0.7.0
+    :tag: v0.7.1
```
bassosimone: @lorenzoPrimi you should probably bump this to be `v0.7.2`.

ooniprobe/AppDelegate.mm
```diff
@@ -114,8 +121,13 @@ - (void)alertView:(UIAlertView *)alertView clickedButtonAtIndex:(NSInteger)butto
     if (buttonIndex == 1 && alertView.tag == 1){
         [self openBrowser];
     }
+    else if (buttonIndex == 1 && alertView.tag == 2){
```
bassosimone: @lorenzoPrimi it would probably help to have `#define`s for these numbers, or it's a bit tricky to follow the code.

ooniprobe/AppDelegate.mm
```diff
@@ -143,6 +155,53 @@ -(void)application:(UIApplication *)application didReceiveLocalNotification:(UIL
     //TODO start your long running bg task here
 }
 
+- (BOOL)application:(UIApplication *)application handleOpenURL:(NSURL *)url {
+    //NSLog(@"url recieved: %@", url);
+    //NSLog(@"query string: %@", [url query]);
+    
+    //creating parameters dict
+    NSDictionary *dict = [DictionaryUtility parseQueryString:[url query]];
+    NSDictionary *parameters = [DictionaryUtility getParametersFromDict:dict];
```
bassosimone: - [ ] @bassosimone needs to check `DictionaryUtility`

ooniprobe/AppDelegate.mm
```diff
@@ -143,6 +155,53 @@ -(void)application:(UIApplication *)application didReceiveLocalNotification:(UIL
     //TODO start your long running bg task here
 }
 
+- (BOOL)application:(UIApplication *)application handleOpenURL:(NSURL *)url {
```
bassosimone: @lorenzoPrimi my understanding is that `handleOpenUrl` is an hardcoded name for iOS. This makes it a little more complex to read the code &mdash; compared to the case where the callback name was more specific &mdash; since I do not immediately understand who calls this function. I believe it would be helpful to have a short documentation (one or two lines) before a function (especially for this fixed-name callbacks) to help the reader :-).

ooniprobe/AppDelegate.mm
```diff
@@ -143,6 +155,53 @@ -(void)application:(UIApplication *)application didReceiveLocalNotification:(UIL
...
+    
+    //creating parameters dict
+    NSDictionary *dict = [DictionaryUtility parseQueryString:[url query]];
+    NSDictionary *parameters = [DictionaryUtility getParametersFromDict:dict];
+    
+    NSString *minimum_version = [parameters objectForKey:@"mv"];
+    if (minimum_version != nil){
+        if ([minimum_version compare:[[[NSBundle mainBundle] infoDictionary] objectForKey:@"CFBundleShortVersionString"] options:NSNumericSearch] == NSOrderedDescending) {
+            //actualVersion is lower than the requiredVersion
+            UIAlertView *alertView = [[UIAlertView alloc] initWithTitle:NSLocalizedString(@"ooniprobe_outdate", nil) message:NSLocalizedString(@"ooniprobe_outdate_msg", nil) delegate:nil cancelButtonTitle:NSLocalizedString(@"Cancel", nil) otherButtonTitles:NSLocalizedString(@"ok", nil), nil];
```
bassosimone: @lorenzoPrimi I was wondering: can `alloc` fail on iOS such that `nil` is returned? What happens in `iOS` if you access a `nil` value? Segmentation fault or is this converted into an "exception" by the framework?

ooniprobe/AppDelegate.mm
```diff
@@ -143,6 +155,53 @@ -(void)application:(UIApplication *)application didReceiveLocalNotification:(UIL
...
+            alertView.tag = 2;
```
bassosimone: (yes, this is probably the tag about which I was asking above; nonetheless, _IMO_ having macros would also make the code more easy to refactor and to keep under control as it grows in size.)

ooniprobe/AppDelegate.mm
bassosimone: Ah, uhm, okay: now I figured that `handleOpenURL` is the method called by iOS on the `UIApplication` when the user clicks on a URL. It was honestly much less difficult than I initially thought (_oops_...). Anyway, maybe adding one line of documentation may be a good idea anyway.

ooniprobe/AppDelegate.mm
```diff
@@ -34,6 +37,10 @@ - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(
     return YES;
 }
 
+- (BOOL)application:(UIApplication *)application continueUserActivity:(NSUserActivity *)userActivity  restorationHandler:(void (^)(NSArray *restorableObjects))restorationHandler{
```
bassosimone: From the [Apple docs](https://developer.apple.com/documentation/uikit/uiapplicationdelegate/1623072-application?preferredLanguage=occ) I read:
> The app calls this method when it receives data associated with a user activity. For example, when the user uses Handoff to transfer an activity from a different device. 

I also read that it has to do with SiriKit. Maybe I've not digged down enough, but it is my understanding that this method returning `YES` is required to continue inside the app the activity initiated clicking in the URL.

Correct?

ooniprobe/AppDelegate.mm
```diff
@@ -143,6 +155,53 @@ -(void)application:(UIApplication *)application didReceiveLocalNotification:(UIL
...
+    NSString *minimum_version = [parameters objectForKey:@"mv"];
+    if (minimum_version != nil){
+        if ([minimum_version compare:[[[NSBundle mainBundle] infoDictionary] objectForKey:@"CFBundleShortVersionString"] options:NSNumericSearch] == NSOrderedDescending) {
+            //actualVersion is lower than the requiredVersion
+            UIAlertView *alertView = [[UIAlertView alloc] initWithTitle:NSLocalizedString(@"ooniprobe_outdate", nil) message:NSLocalizedString(@"ooniprobe_outdate_msg", nil) delegate:nil cancelButtonTitle:NSLocalizedString(@"Cancel", nil) otherButtonTitles:NSLocalizedString(@"ok", nil), nil];
+            alertView.tag = 2;
+            [alertView show];
+        }
+        else {
+            NSString *action = [url host];
```
bassosimone: @lorenzoPrimi is it correct to state that here you're taking the `host` of the URL because the URL starts like `ooni://nettest`? (Maybe it would help a comment on top of the routine stating what the URL format is, or it would also possibly do a link to the specification of the URLs)

ooniprobe/AppDelegate.mm
```diff
@@ -143,6 +155,53 @@ -(void)application:(UIApplication *)application didReceiveLocalNotification:(UIL
...
+            if ([action isEqualToString:@"nettest"]){
+                //For now checking only test name
+                if ([parameters objectForKey:@"tn"] && [[Tests currentTests] getTestWithName:[parameters objectForKey:@"tn"]])
+                    [self openURIschemeScreen:parameters];
+                else {
+                    UIAlertView *alertView = [[UIAlertView alloc] initWithTitle:NSLocalizedString(@"invalid_parameter", nil) message:[NSString stringWithFormat:@"%@ : %@", NSLocalizedString(@"test_name", nil), [parameters objectForKey:@"tn"]] delegate:nil cancelButtonTitle:NSLocalizedString(@"Cancel", nil) otherButtonTitles:nil];
```
bassosimone: For reading the diffs (especially in the terminal) it may be better to wrap very long lines.

ooniprobe/AppDelegate.mm
```diff
@@ -143,6 +155,53 @@ -(void)application:(UIApplication *)application didReceiveLocalNotification:(UIL
...
+                }
+            }
+        }
+    }
+    return YES;
+}
+
+    
+-(void)openURIschemeScreen:(NSDictionary*)parameters{
+    dispatch_async(dispatch_get_main_queue(), ^{
```
bassosimone: BTW, I really _love_ Apple's API for dispatching async tasks.

ooniprobe/AppDelegate.mm
```diff
@@ -143,6 +155,53 @@ -(void)application:(UIApplication *)application didReceiveLocalNotification:(UIL
...
+        UIStoryboard *mainStoryboard = [UIStoryboard storyboardWithName:@"Main" bundle: nil];
+        UINavigationController *nvc = [mainStoryboard instantiateViewControllerWithIdentifier:@"runtestNC"];
+        RunTestViewController *rvc = (RunTestViewController*)[nvc.viewControllers objectAtIndex:0];
```
bassosimone: - [ ] @bassosimone needs to read carefully `RunTestViewController`

ooniprobe/AppDelegate.mm
```diff
@@ -143,6 +155,53 @@ -(void)application:(UIApplication *)application didReceiveLocalNotification:(UIL
...
+        if ([minimum_version compare:[[[NSBundle mainBundle] infoDictionary] objectForKey:@"CFBundleShortVersionString"] options:NSNumericSearch] == NSOrderedDescending) {
+            //actualVersion is lower than the requiredVersion
+            UIAlertView *alertView = [[UIAlertView alloc] initWithTitle:NSLocalizedString(@"ooniprobe_outdate", nil) message:NSLocalizedString(@"ooniprobe_outdate_msg", nil) delegate:nil cancelButtonTitle:NSLocalizedString(@"Cancel", nil) otherButtonTitles:NSLocalizedString(@"ok", nil), nil];
+            alertView.tag = 2;
+            [alertView show];
+        }
+        else {
+            NSString *action = [url host];
+            if ([action isEqualToString:@"nettest"]){
+                //For now checking only test name
```
bassosimone: Does this mean that we should _check_ for more parameters or that for now the code is limited in that it does not actually run the test but just makes sure that the test name is okay?

ooniprobe/AppDelegate.mm
bassosimone: BTW, @lorenzoPrimi, is it correct that you need to run this in the `I/O` queue context so to access the view?

ooniprobe/Utility/DictionaryUtility.h
```diff
@@ -0,0 +1,15 @@
+//
+//  DictionaryUtility.h
+//  ooniprobe
+//
+//  Created by Lorenzo Primiterra on 06/08/17.
+//  Copyright © 2017 Simone Basso. All rights reserved.
+//
```
bassosimone: Can you update the comment? (The presence of this comment, strictly speaking, make ooniprobe-ios non FLOSS)

ooniprobe/Utility/DictionaryUtility.m
```diff
@@ -0,0 +1,47 @@
+//
+//  DictionaryUtility.m
+//  ooniprobe
+//
+//  Created by Lorenzo Primiterra on 06/08/17.
+//  Copyright © 2017 Simone Basso. All rights reserved.
```
bassosimone: Ditto

ooniprobe/Utility/DictionaryUtility.m
```diff
@@ -0,0 +1,47 @@
...
+        NSData *data = [current dataUsingEncoding:NSUTF8StringEncoding];
+        NSDictionary *dictionary = [NSJSONSerialization JSONObjectWithData:data options:kNilOptions error:&error];
+        if (error == nil){
+            [parameters setObject:dictionary forKey:key];
+            //NSLog(@"json dictionary for key %@ : %@", key, dictionary);
+        }
+        else{
+            [parameters setObject:current forKey:key];
+            //NSLog(@"json object for key %@ : %@", key, current);
+            //NSLog(@"error: %@", error);
```
bassosimone: So, this basically means: walk over all the dict, if the parameter is a JSON, unserialize it, otherwise just leave it as is?

ooniprobe/View/PastTestsViewController.mm
```diff
@@ -109,6 +109,10 @@ - (NSInteger) tableView:(UITableView *)tableView numberOfRowsInSection:(NSIntege
     return [finishedTests count];
 }
 
+- (CGFloat) tableView:(UITableView *)tableView heightForHeaderInSection:(NSInteger)section
+{
+    return CGFLOAT_MIN;
+}
```
bassosimone: Is this a style fix?

ooniprobe/View/ResultSelectorViewController.mm
```diff
@@ -43,6 +43,10 @@ - (NSInteger)tableView:(UITableView *)tableView numberOfRowsInSection:(NSInteger
     return [testResults count];
 }
 
+- (CGFloat) tableView:(UITableView *)tableView heightForHeaderInSection:(NSInteger)section
+{
+    return CGFLOAT_MIN;
+}
```
bassosimone: Is this a style fix?

ooniprobe/View/RunTestViewController.h
```diff
@@ -0,0 +1,33 @@
+//
+//  RunTestTableViewController.h
+//  ooniprobe
+//
+//  Created by Lorenzo Primiterra on 06/08/17.
+//  Copyright © 2017 Simone Basso. All rights reserved.
+//
```
bassosimone: Same comment about the copyright

bassosimone: (Uhm, maybe we can change something in the Xcode config so to get it by default?)

ooniprobe/View/RunTestViewController.mm
```diff
@@ -0,0 +1,121 @@
+//
+//  RunTestTableViewController.m
+//  ooniprobe
+//
+//  Created by Lorenzo Primiterra on 06/08/17.
+//  Copyright © 2017 Simone Basso. All rights reserved.
+//
```
bassosimone: Ditto

ooniprobe/View/RunTestViewController.mm
```diff
@@ -0,0 +1,121 @@
...
+
+#import "RunTestViewController.h"
+
+@interface RunTestViewController ()
```
bassosimone: @lorenzoPrimi I have seen this pattern frequently used, but I don't fully understand it. Why one needs to repeat the interface again in the implementation file just leaving it empty?

ooniprobe/View/RunTestsViewController.mm
```diff
@@ -75,6 +76,8 @@ - (IBAction)runTests:(id)sender event:(id)event {
     CGPoint currentTouchPosition = [[[event allTouches] anyObject] locationInView:self.tableView];
     NSIndexPath *indexPath = [self.tableView indexPathForRowAtPoint: currentTouchPosition];
     NetworkMeasurement *current = [currentTests.availableNetworkMeasurements objectAtIndex:indexPath.row];
+    //TODO set test list
+    current.inputs = nil;
```
bassosimone: Does this mean that we still need to do work to finish this release, or is this in preparation for future changes?

ooniprobe/View/SettingsTableViewController.mm
```diff
@@ -81,6 +81,12 @@ - (NSString *)tableView:(UITableView *)tableView titleForHeaderInSection:(NSInte
     return NSLocalizedString(@"advanced_settings", nil);
 }
 
+-(void)tableView:(UITableView *)tableView willDisplayHeaderView:(UIView *)view forSection:(NSInteger)section
+{
+    UITableViewHeaderFooterView *header = (UITableViewHeaderFooterView *)view;
+    header.textLabel.font = [UIFont fontWithName:@"FiraSansOT-Bold" size:15];
+    header.textLabel.textColor = color_off_black;
+}
```
bassosimone: This is a style fix, right?

ooniprobe/View/SettingsTableViewController.mm
```diff
@@ -224,8 +230,8 @@ - (void)tableView:(UITableView *)tableView didSelectRowAtIndexPath:(NSIndexPath
 {
     if (indexPath.section == 2 && indexPath.row == [advancedItems count] -1){
         NSString *current = [advancedItems objectAtIndex:indexPath.row];
-        UIAlertView *alert = [[UIAlertView alloc] initWithTitle:NSLocalizedString(current, @"") message:nil delegate:self cancelButtonTitle:NSLocalizedString(@"cancel", @"") otherButtonTitles:NSLocalizedString(@"ok", nil), nil];
-        alert.tag = indexPath.row;
+        UIAlertView *alert = [[UIAlertView alloc] initWithTitle:NSLocalizedString(current, @"") message:nil delegate:self cancelButtonTitle:NSLocalizedString(@"cancel", @"") otherButtonTitles:NSLocalizedString(@"ok", nil), NSLocalizedString(@"set_default", nil), nil];
+        alert.tag = 1;
```
bassosimone: - [ ] @bassosimone needs to understand this change

ooniprobe/ooniprobe.entitlements
```diff
@@ -4,5 +4,9 @@
 <dict>
 	<key>aps-environment</key>
 	<string>development</string>
+	<key>com.apple.developer.associated-domains</key>
+	<array>
+		<string>applinks:run.ooni.org</string>
```
bassosimone: Isn't it `run.ooni.io`?

ooniprobe/AppDelegate.mm
bassosimone: Uhm, from [run.ooni.io](https://run.ooni.io/) I _also_ see URIs like: `https://run.ooni.io/nettest?tn=ndt&mv=1.2.0`.

bassosimone: The point is: there the host is `run.ooni.io`.

ooniprobe.xcodeproj/project.pbxproj
```diff
@@ -744,12 +762,12 @@
...
 				);
 				INFOPLIST_FILE = ooniprobe/Info.plist;
-				IPHONEOS_DEPLOYMENT_TARGET = 8.0;
+				IPHONEOS_DEPLOYMENT_TARGET = 9.0;
 				LD_RUNPATH_SEARCH_PATHS = "$(inherited)";
 				LIBRARY_SEARCH_PATHS = "$(inherited)";
 				PRODUCT_BUNDLE_IDENTIFIER = org.openobservatory.ooniprobe;
 				PRODUCT_NAME = ooniprobe;
-				PROVISIONING_PROFILE = "9c5b2b6c-065a-4427-96fc-1cba88a959e8";
+				PROVISIONING_PROFILE = "de5a74e0-ac7a-48a3-b51c-fa0b65ad0759";
```
bassosimone: @lorenzoPrimi do you know why this changed?

ooniprobe/AppDelegate.mm
bassosimone: It looks good!

ooniprobe/AppDelegate.mm
bassosimone: _Question_: how do you register handling URLs with iOS? I assume with the entitlements change. If this is correct, I think I have missed the place where `ooni://...` URLs are registered.

If my additional guess that here you will handle both `ooni://...` and `https://run.ooni.io/...` links, then I believe the code should be updated to make sure it handles both cases.

Besides, maybe, checking at the beginning that the URL either starts with `ooni://...` or `https://run.ooni.io/...` _may_ be useful in future to detect possible errors that we could make where we accidentally pass to this function _other_ URLs.

ooniprobe/PrefixHeader.pch
```diff
@@ -10,6 +10,8 @@
 // Include any system framework and library headers here that should be included in all compilation units.
 // You will also need to set the Prefix Header build setting of one or more of your targets to reference this file.
 
+#define COLLECTOR_ADDRESS @"https://b.collector.ooni.io"
+
```
bassosimone: I wonder what change in Xcode was such that this ended up here 🤔

ooniprobe/View/RunTestViewController.mm
```diff
@@ -0,0 +1,121 @@
...
+@synthesize testName, testArguments, testDecription;
+
+- (void)viewDidLoad {
+    [super viewDidLoad];
+    [[NSNotificationCenter defaultCenter] addObserver:self selector:@selector(reloadTestUI) name:@"reloadTable" object:nil];
+
+    currentTests = [Tests currentTests];
+    [self.runButton setTitle:[NSString stringWithFormat:@"   %@   ", NSLocalizedString(@"run", nil)] forState:UIControlStateNormal];
+
+    if (testDecription != nil)
```
bassosimone: This is conditionally set using a `synthetized` setter in the code above, correct?

ooniprobe/View/RunTestViewController.mm
```diff
@@ -0,0 +1,121 @@
...
+}
+
+-(IBAction)run:(id)sender {
+    NetworkMeasurement *current = [currentTests getTestWithName:testName];
+    if ([urls count] > 0){
+        current.inputs = urls;
+    }
+    [current run];
+    [self dismissViewControllerAnimated:TRUE completion:^{
+        [[NSNotificationCenter defaultCenter] postNotificationName:@"reloadTable" object:nil];
```
bassosimone: So, does this mean that, when the view is closed, you post the `@"reloadTable"` message that is the one you use to reload, if I understand correctly, the `run test` view and, possibly, also the `past test results` view?

ooniprobe/View/RunTestViewController.mm
```diff
@@ -0,0 +1,121 @@
...
+@interface RunTestViewController ()
+
+@end
+
+@implementation RunTestViewController
+@synthesize testName, testArguments, testDecription;
+
+- (void)viewDidLoad {
+    [super viewDidLoad];
+    [[NSNotificationCenter defaultCenter] addObserver:self selector:@selector(reloadTestUI) name:@"reloadTable" object:nil];
```
bassosimone: This basically register a receiver for `@"reloadTable"` that is the notification you emit, among other things, at the end of the test, correct? This means that, when the test ends, the current view would update, isn't it?

ooniprobe/View/RunTestViewController.mm
bassosimone: Is the rationale of posting this message that you will then be able to update the other views since you basically have started the test from another code-path and view?

ooniprobe/View/RunTestsViewController.mm
```diff
@@ -137,6 +145,12 @@ -(void)showToast{
     [self.view makeToast:NSLocalizedString(@"ooniprobe_configured", nil) duration:3.0 position:CSToastPositionBottom style:style];
 }
 
+-(void)showToast:(NSNotification *) notification{
```
bassosimone: Since my understanding is that this is the toast showed at end of test, maybe you can call it `showToastTestFinished`?

bassosimone: Oh, I see that name already exists. This confused me, probably: what is the purpose of this toast?

bassosimone: mmm, `git grep shotToastTestFinished` does not return any row tho.

bassosimone: Disregard, now I understand. The `showToast` is the generic show toast function (maybe rename it `showToastGeneric`?) and the one above is specific for configuration. The `@"showToastTestFinished"` is just the name of a specific notification that you're using inside the app (i.e. it is a string, not a function).

ooniprobe/View/SettingsTableViewController.mm
bassosimone: So, it seems `indexPath.row` was unused and now you're using a stable tag such that you know that when the tag is equal to one you have clicked to modify the collector?

bassosimone:
> @lorenzoPrimi in the future can you keep the changes related to code separate from the ones just updating translations (that should maybe only happen before the release is ready to be done)? It makes reading the diff a bit easier.

I second this. I'd suggest we create a separate PR for translations that shall be merged quickly (one just needs to compile _once_ the app in Xcode to see there is no evident breakage). Then, github offers a button to merge back master into the open PR. This allows you, basically, to bring into your currently open PR the changes you did meanwhile on master (i.e. you see all of them when you test).

ooniprobe/AppDelegate.mm
lorenzoPrimi: Everything is the Info.plist file.
The links on http are handled by this other method

- (BOOL)application:(UIApplication *)application continueUserActivity:(NSUserActivity *)userActivity  restorationHandler:(void (^)(NSArray *restorableObjects))restorationHandler{

ooniprobe/AppDelegate.mm
lorenzoPrimi: if you try to access a parameter into a var that is nil it crashes the app

ooniprobe/AppDelegate.mm
lorenzoPrimi: Answered above. We only handle https://run.ooni.io here

ooniprobe/AppDelegate.mm
lorenzoPrimi: Here the host is run.ooni.io when opened from the website, is nettest when opened from a link ooni://nettest.

ooniprobe/AppDelegate.mm
lorenzoPrimi: - The call to dispatch_async is performed from the main thread.
- It's mainly used for UI updating threads

ooniprobe/AppDelegate.mm
lorenzoPrimi: We only have test name to check for now

ooniprobe/Utility/DictionaryUtility.m
lorenzoPrimi: This is to distinguish between subparameters that can be strings or other json nested.

ooniprobe/View/PastTestsViewController.mm
lorenzoPrimi: Yes, to remove headers

ooniprobe/View/RunTestViewController.h
lorenzoPrimi: If you know this settings let me know !

ooniprobe/View/RunTestViewController.mm
lorenzoPrimi: It's the default pattern when you create a new class, I never asked myself why

ooniprobe/View/RunTestsViewController.mm
lorenzoPrimi: The second thing you said

ooniprobe/View/SettingsTableViewController.mm
lorenzoPrimi: Yes

ooniprobe/View/SettingsTableViewController.mm
lorenzoPrimi: That was a mistake, there was no delagate handling. Now there is only one delegate, to set the default collector.

ooniprobe/ooniprobe.entitlements
lorenzoPrimi: It is, are you looking at the last commit?

ooniprobe.xcodeproj/project.pbxproj
lorenzoPrimi: new provisioning sent to me by nick

ooniprobe/PrefixHeader.pch
lorenzoPrimi: See above. There is an option in the settings to set the default collector in case you mess with the settings

ooniprobe/View/RunTestViewController.mm
lorenzoPrimi: I don't get your comment

ooniprobe/View/RunTestViewController.mm
lorenzoPrimi: Yes, because I will post this message after dismissing this view controller

ooniprobe/View/RunTestViewController.mm
lorenzoPrimi: yes

ChangeLog.md
```diff
@@ -1,8 +1,28 @@
...
+
+New features:
+
+* Added support for running tests with custom URI scheme
+
+* Added possibility to open links from run.ooni.io
+
+* Drop support for iOS < 9
+
+measurement-kit version: 0.7.3
```
bassosimone: It's actually `v0.7.2`

ooniprobe/AppDelegate.mm
bassosimone: Okay, maybe can you add comments saying that? It simplifies reading the code.

ooniprobe/AppDelegate.mm
bassosimone: okay, and can `alloc` fail?

ooniprobe/PrefixHeader.pch
bassosimone: Okay

ooniprobe/Utility/DictionaryUtility.m
bassosimone: Maybe you should add a comment so the code is more clear

ooniprobe/View/PastTestsViewController.mm
bassosimone: Okay, thanks!

ooniprobe/View/RunTestViewController.mm
bassosimone: Thanks

ooniprobe/View/RunTestViewController.mm
bassosimone: Nevermind; beware that there is a typo: it should be `testDescription` not `testDecription`

ooniprobe/View/RunTestViewController.mm
bassosimone: Okay, thanks

ooniprobe/View/RunTestsViewController.mm
bassosimone: Ok

ooniprobe/AppDelegate.mm
lorenzoPrimi: which alloc?

hellais:
Can we get travis to become happy? I think it's just a matter of removing the `brew install gpg2` line:

```
$ brew install gpg2
Error: gnupg 1.4.21 is already installed
To upgrade to 2.2.1, run `brew upgrade gnupg`
```

ooniprobe/AppDelegate.mm
bassosimone: `[UIAlertView alloc]`

ooniprobe/AppDelegate.mm
bassosimone: @lorenzoPrimi do you know whether it's possible to configure Xcode to reformat the code (or can we use directly `clang-format` for such purpose)? After this release cycle I'd like to reformat the code a bit to have shorter lines.

ooniprobe/Utility/DictionaryUtility.h
bassosimone: ^^^ @lorenzoPrimi

ooniprobe/Utility/DictionaryUtility.m
bassosimone: ^^^ @lorenzoPrimi

ooniprobe/View/BrowserViewController.m
```diff
@@ -1,10 +1,6 @@
-//
-//  BrowserViewController.m
-//  ooniprobe
-//
-//  Created by Lorenzo Primiterra on 23/06/17.
-//  Copyright © 2017 Simone Basso. All rights reserved.
-//
+// Part of MeasurementKit <https://measurement-kit.github.io/>.
+// MeasurementKit is free software. See AUTHORS and LICENSE for more
+// information on the copying conditions.
```
bassosimone: Note to self: after this release cycle, these comments should actually be updated to reference OONI. Since I already have a script for dealing with that in MK, I can apply it also to ooniprobe-ios and -android.

ooniprobe/View/RunTestsViewController.mm
bassosimone: ^^^ @lorenzoPrimi is it correct what I am saying here?

ooniprobe/AppDelegate.mm
lorenzoPrimi: no idea. I don't even think it can happen such scenario :D

ooniprobe/View/RunTestsViewController.mm
lorenzoPrimi: Yes

ooniprobe/Utility/VersionUtility.mm
```diff
@@ -0,0 +1,16 @@
+//
+//  VersionUtility.m
+//  ooniprobe
+//
+//  Created by Lorenzo Primiterra on 24/09/17.
+//  Copyright © 2017 Simone Basso. All rights reserved.
```
hellais: This should be the standard MK header

ooniprobe/Utility/VersionUtility.mm
```diff
@@ -0,0 +1,16 @@
...
+//
+
+#import "VersionUtility.h"
+
+@implementation VersionUtility
+
++ (NSString*)get_software_version{
+    return [NSString stringWithFormat:@"%@%@+%@", [[[NSBundle mainBundle] infoDictionary] objectForKey:@"CFBundleShortVersionString"], release_name, [[[NSBundle mainBundle] infoDictionary] objectForKey:@"CFBundleVersion"]];
```
hellais: Isn't there missing a `-` in between?

hellais: Ok I now see the release_version contains the `-`

ooniprobe/Utility/VersionUtility.mm
hellais: Or just not be there as for the rest of the code.

bassosimone: It should probably mention `OONI Probe` not `MK`. TBH, I am also fine it removing all headers if that's what you like the most :-)

ooniprobe/AppDelegate.mm
bassosimone: I tried to dig in a little bit. What I have found out (but this is still cursory) is the following:

1. ["Objective-C safely handles sending messages to nil by ignoring them."](https://www.quora.com/Why-do-I-need-to-do-Class-*obj-Class-alloc-init-on-one-line/answer/Andy-Dent), which means that `Foo *foo = [[Foo alloc] init]` is safe

2. [that there are specific patterns that you should follow in the init handler](https://stackoverflow.com/a/10358995)

bassosimone: So, in short: the scenario _can_ happen but you don't need to worry about it (at least in this context) because attempting to send messages to `nil` is something supported by ObjC.

ooniprobe/AppDelegate.mm
bassosimone: I checked, there is clearly support for objc in `clang-format`. I think we _should_ really adopt it.

hellais:
Yay, travis is happy: https://travis-ci.org/TheTorProject/ooniprobe-ios/builds/279316255?utm_source=github_status&utm_medium=notification 🎉

PR #135, opened 2017-09-26
hotfix(VersionUtility): do not include build number

bassosimone:
This breaks OONI Backend.

See TheTorProject/ooni-backend#111 and TheTorProject/ooni-sysadmin#164.


bassosimone:
Will merge as soon as Travis is happy.

bassosimone:
Ah, right, this build will take more or less forever because Travis has many more iOS and macOS builds than initially anticipated and is now trying various techniques to cope with that.

bassosimone:
Okay, I am going to merge this without waiting for travis.

PR #141, opened 2017-09-28
Fixing #132

### LINKED ISSUE #132: The run.ooni.io view persists if not dismissed
As I wrote on Slack:

> if I click on a link, then I don't run a test, then I click on a second link, I still see the running screen of the former test, rather than the running screen of the second one, as I would expect

lorenzoPrimi:
Closes #132

### DIFF
--- a/ooniprobe/AppDelegate.mm
+++ b/ooniprobe/AppDelegate.mm
@@ -221,9 +221,10 @@
         if ([parameters objectForKey:@"td"])
             [rvc setTestDescription:[parameters objectForKey:@"td"]];
         if (self.window.rootViewController.view.window != nil)
+            //only main view controller is visible
             [self.window.rootViewController presentViewController:nvc animated:YES completion:nil];
         else {
-            //view is not in the window hierarchy
+            //main view controller is not in the window hierarchy, so overlay window was presented already, reloading parameters
             [[NSNotificationCenter defaultCenter] postNotificationName:@"reloadTest" object:nil userInfo:parameters];
         }
     });
--- a/ooniprobe/AppDelegate.mm
+++ b/ooniprobe/AppDelegate.mm
@@ -220,7 +220,12 @@
             [rvc setTestArguments:[parameters objectForKey:@"ta"]];
         if ([parameters objectForKey:@"td"])
             [rvc setTestDescription:[parameters objectForKey:@"td"]];
-        [self.window.rootViewController presentViewController:nvc animated:YES completion:nil];
+        if (self.window.rootViewController.view.window != nil)
+            [self.window.rootViewController presentViewController:nvc animated:YES completion:nil];
+        else {
+            //view is not in the window hierarchy
+            [[NSNotificationCenter defaultCenter] postNotificationName:@"reloadTest" object:nil userInfo:parameters];
+        }
     });
 }
 
--- a/ooniprobe/View/RunTestViewController.mm
+++ b/ooniprobe/View/RunTestViewController.mm
@@ -17,7 +17,7 @@
 -(void)configureView{
     currentTests = [Tests currentTests];
     [self.runButton setTitle:[NSString stringWithFormat:@"   %@   ", NSLocalizedString(@"run", nil)] forState:UIControlStateNormal];
-    
+
     if (testDescription != nil)
         [self.titleLabel setText:[NSString stringWithFormat:@"%@", testDescription]];
     else
--- a/ooniprobe/View/RunTestViewController.mm
+++ b/ooniprobe/View/RunTestViewController.mm
@@ -10,15 +10,19 @@
 - (void)viewDidLoad {
     [super viewDidLoad];
     [[NSNotificationCenter defaultCenter] addObserver:self selector:@selector(reloadTestUI) name:@"reloadTable" object:nil];
+    [[NSNotificationCenter defaultCenter] addObserver:self selector:@selector(reloadTest:) name:@"reloadTest" object:nil];
+    [self configureView];
+}
 
+-(void)configureView{
     currentTests = [Tests currentTests];
     [self.runButton setTitle:[NSString stringWithFormat:@"   %@   ", NSLocalizedString(@"run", nil)] forState:UIControlStateNormal];
-
+    
     if (testDescription != nil)
         [self.titleLabel setText:[NSString stringWithFormat:@"%@", testDescription]];
     else
         [self.titleLabel setText:NSLocalizedString(@"run_test_message", nil)];
-
+    
     [self.test_detailsLabel setText:NSLocalizedString(@"test_details", nil)];
     NetworkMeasurement *current = [currentTests getTestWithName:testName];
     [self.test_titleLabel setText:NSLocalizedString(current.name, nil)];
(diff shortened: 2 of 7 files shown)


ooniprobe/View/RunTestViewController.mm
```diff
@@ -10,23 +10,28 @@ @implementation RunTestViewController
...
     [self.test_detailsLabel setText:NSLocalizedString(@"test_details", nil)];
     NetworkMeasurement *current = [currentTests getTestWithName:testName];
     [self.test_titleLabel setText:NSLocalizedString(current.name, nil)];
     [self.test_iconImage setImage:[UIImage imageNamed:[NSString stringWithFormat:@"%@_big", current.name]]];
     
     self.tableView.estimatedRowHeight = 44.0;
     self.tableView.rowHeight = UITableViewAutomaticDimension;
-
+    
+    rows = [[NSArray alloc] init];
```
hellais: How come this variable was not defined previously? Is this what the fix is? That basically you were re-using every time the same variable?

lorenzoPrimi: It's initialized below https://github.com/TheTorProject/ooniprobe-ios/blob/e858d8efb75c55b6854d9860f1bdf810c54e5be2/ooniprobe/View/RunTestViewController.mm#L38

This is needed to "reset" the variable when the screen is reopened without closing it.

bassosimone:
Also this diff looks reasonable, but I need to further test it before signing it off.

ooniprobe/View/RunTestViewController.mm
```diff
@@ -10,23 +10,28 @@ @implementation RunTestViewController
...
     [[NSNotificationCenter defaultCenter] addObserver:self selector:@selector(reloadTestUI) name:@"reloadTable" object:nil];
+    [[NSNotificationCenter defaultCenter] addObserver:self selector:@selector(reloadTest:) name:@"reloadTest" object:nil];
+    [self configureView];
+}
 
+-(void)configureView{
     currentTests = [Tests currentTests];
     [self.runButton setTitle:[NSString stringWithFormat:@"   %@   ", NSLocalizedString(@"run", nil)] forState:UIControlStateNormal];
-
+
```
bassosimone: @lorenzoPrimi nit: you should get rid of this spaces-only lines.

ooniprobe/View/RunTestViewController.mm
bassosimone: @lorenzoPrimi maybe it helps a comment explaining why you're re-initializing it.

ooniprobe/View/RunTestViewController.mm
```diff
@@ -38,7 +43,9 @@ - (void)viewDidLoad {
     else {
         [self.tableView setSeparatorStyle:UITableViewCellSeparatorStyleNone];
     }
-    
+    dispatch_async(dispatch_get_main_queue(), ^{
+        [self.tableView reloadData];
+    });
```
bassosimone: I don't get exactly why this is necessary now and was not necessary before.

ooniprobe/View/RunTestViewController.mm
lorenzoPrimi: to be sure that the rows var is resetted when I open a new screen over the current one

ooniprobe/View/RunTestViewController.mm
lorenzoPrimi: Now XCode throws a warning, before didn't :D

ooniprobe/AppDelegate.mm
```diff
@@ -220,7 +220,12 @@ -(void)openURIschemeScreen:(NSDictionary*)parameters{
             [rvc setTestArguments:[parameters objectForKey:@"ta"]];
         if ([parameters objectForKey:@"td"])
             [rvc setTestDescription:[parameters objectForKey:@"td"]];
-        [self.window.rootViewController presentViewController:nvc animated:YES completion:nil];
+        if (self.window.rootViewController.view.window != nil)
+            [self.window.rootViewController presentViewController:nvc animated:YES completion:nil];
+        else {
+            //view is not in the window hierarchy
+            [[NSNotificationCenter defaultCenter] postNotificationName:@"reloadTest" object:nil userInfo:parameters];
+        }
```
bassosimone: Which is the case where you have already created a view and you're just recycling it? The second?

lorenzoPrimi: yes

bassosimone: Okay, but why would the `.window` be `!= nil` in that case?

ooniprobe/View/RunTestViewController.mm
bassosimone: Yes yes, that was clear to me. I was suggesting to put this into a comment.

ooniprobe/View/RunTestViewController.mm
bassosimone: Yeah, but before you were not calling `reloadData` at all, no?

lorenzoPrimi: Oh I thought the comment was about the dispatch_async.
No before I wasn't because I wasn't handling the case of reloading the view with new parameters.

ooniprobe/AppDelegate.mm
lorenzoPrimi: In this case view.window is the ViewController of the main app (the one with run tests etc) not the overlay view.
It's nil when is covered. Or at least it's what I learnt from stackoverflow

ooniprobe/View/RunTestViewController.mm
bassosimone: oh, right, thanks for explaining!

ooniprobe/AppDelegate.mm
bassosimone: ack!

bassosimone:
@lorenzoPrimi I guess this is the next PR I'll take care of. I'll also deal with conflicts right away.

ooniprobe/AppDelegate.mm
bassosimone: Can you maybe add a comment saying that and referencing the stack-overflow link?

lorenzoPrimi: https://stackoverflow.com/questions/26022756/warning-attempt-to-present-on-whose-view-is-not-in-the-window-hierarchy-s

https://stackoverflow.com/questions/30941424/is-myviewcontroller-in-the-window-hierarchy

bassosimone: I meant, a comment _in the code_

bassosimone: Like here: https://github.com/TheTorProject/ooniprobe-ios/pull/141/files#diff-eff536c28a49eda653581c9cd5605e23R223

ooniprobe/View/RunTestViewController.mm
bassosimone: Again, I meant a comment in the code, just above when you reinit the array

ooniprobe/View/RunTestViewController.mm
bassosimone: Again, maybe a comment in the code may help.

ooniprobe/AppDelegate.mm
lorenzoPrimi: I don't like putting links in the code, expecially from random stackoverflow answers and when what the code is doing is so clear. 
Can I just explain when the view is nill?

PR #146, opened 2017-10-15
Setting devicetoken empty when null

lorenzoPrimi:


ooniprobe/AppDelegate.mm
```diff
@@ -27,6 +27,9 @@ - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(
 
     [[UIApplication sharedApplication] setStatusBarStyle:UIStatusBarStyleLightContent];
     
+    
+    //TODO register as a probe if I have never done it. This is needed to register the probe even if push notifications are turned off.
```
lorenzoPrimi: I still don't know how to handle this. I need a way to understand if the user registered at least once to the notification.

lorenzoPrimi:
Not yet ready for review

lorenzoPrimi:
@bassosimone I've had to duplicate the code for register / update the probe.
This is needed to don't call update on start when the probe is already registered.
Is this the correct way or you have a better idea?

ooniprobe/AppDelegate.mm
```diff
@@ -23,10 +23,14 @@ - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(
     #ifdef RELEASE
     CrashlyticsKit.delegate = self;
     [Fabric with:@[[Crashlytics class]]];
+    [[NotificationService sharedNotificationService] registerProbe];
```
bassosimone: So, this basically means that, whenever the app is loaded, we unconditionally register probe?

ooniprobe/AppDelegate.mm
bassosimone: I think what we really need is some semantic like get-if-modified-since in the backend. Except that here it would be something like register-if-not-already-registered. Otherwise this is going to create many headaches for us.

Makes sense?

bassosimone: To be more clear: it is my understanding that the current API is not idempotent in that you cannot ask to register a probe if it's not already registered and consider that an update otherwise. I'd really like to have a method with this semantic so that we can just issue such an idempotent request on load. This is particularly useful to us because there are cases in which some devices are used both for testing and for production. And, in general, having an idempotent register method is gonna be useful to us. I wonder whether like this it's more clear, @lorenzoPrimi what do you think?

ooniprobe/AppDelegate.mm
lorenzoPrimi: No, it will call the registerprobe function, that will exit if the probe is already registered. 
All my doubts are in the registerProbe function

ooniprobe/Utility/NotificationService.mm
```diff
@@ -124,4 +123,59 @@ -(NSString*)make_path {
...
+    client.supported_tests = supported_tests_list;
+    client.network_type = [network_type UTF8String];
+    client.language = [language UTF8String];
+    client.registry_url = [NOTIFICATION_SERVER UTF8String];
+    
+    if (device_token == nil)
+        device_token = @"";
+    client.device_token = [device_token UTF8String];
+    
+    //Can I directly call auth.load here without calling client.find_location?
```
lorenzoPrimi: @bassosimone

PR #151, opened 2017-10-31
Refactor/NetworkMeasurement

lorenzoPrimi:


lorenzoPrimi:
Should we remove setup_idempotent() and keep it in this way?
`test.set_verbosity(VERBOSITY);
    test.on_log([self](uint32_t type, const char *s) {
#ifdef DEBUG
        NSLog(@“%s”, s);
#endif
    });`

ooniprobe/Model/NetworkMeasurement.mm
```diff
@@ -17,21 +17,6 @@
...
-        mk::set_verbosity(VERBOSITY);
-        mk::on_log([](uint32_t, const char *s) {
-            #ifdef DEBUG
-            NSLog(@"%s", s);
-            #endif
-        });
-        initialized = true;
-    }
-}
-
```
bassosimone: 👍

ooniprobe/Model/NetworkMeasurement.mm
```diff
@@ -441,19 +230,10 @@ -(void)run{
 }
 
 -(void) run_test {
-    [super run_test];
     NSBundle *bundle = [NSBundle mainBundle];
     NSString *path = [bundle pathForResource:@"global" ofType:@"txt"];
```
bassosimone: Are these two variables used?

ooniprobe/Model/NetworkMeasurement.mm
```diff
@@ -503,33 +311,38 @@ -(void)run{
...
-        })
-        .on_entry([self](std::string s) {
-            [self on_entry_ndt:s.c_str()];
-        })
-        .start([self]() {
-            [self testEnded];
-        });
+    //[super run_test];
+    mk::nettests::NdtTest test;
+    [super run_test:test];
```
bassosimone: I would not call `run_test` the `super` method, because it does not actually run the test. I'd call it `init_common` or something that conveys the meaning that that is the common initialization path.

ooniprobe/Model/NetworkMeasurement.mm
```diff
@@ -552,36 +365,52 @@ -(void)run{
...
-    })
-    .on_entry([self](std::string s) {
-        [self on_entry_hhfm:s.c_str()];
-    })
-    .start([self]() {
+    mk::nettests::HttpHeaderFieldManipulationTest test;
+    [super run_test:test];
+    test.on_entry([self](std::string s) {
+        [self on_entry:s.c_str()];
+    });
```
bassosimone: It seems to me that `on_entry` is equal for all tests. Cannot you put this into the common code. If I understand OO correctly, calling `on_entry` in a parent class will resolve to the `on_entry` of the real class.

bassosimone: Otherwise, you can perhaps pass the pointer to the method to the super class common-init method?

ooniprobe/Model/NetworkMeasurement.mm
```diff
@@ -604,37 +433,39 @@ -(void)run{
...
         [self testEnded];
     });
 }
 
+/*
+ Note: here we're reusing NDT function since the entry has
+ basically the same characteristics, as far as deciding the
+ color of the test line is concerned.
+ if the "failure" key exists and is not null then anomaly will be set to 1 (orange)
+ */
```
bassosimone: If it's really equal, wouldn't it be possible to really invoke the same method? (I may be missing come ctx here)

ooniprobe/Model/NetworkMeasurement.mm
lorenzoPrimi: yes, they are removed in the other branch

ooniprobe/Model/NetworkMeasurement.mm
lorenzoPrimi: Only if we move it in the super class. I want to keep class specific methods. Nevermind the algorithm is the same.

ooniprobe/Model/NetworkMeasurement.mm
lorenzoPrimi: The super class won't know that there is an on_entry in the bottom class

PR #165, opened 2018-01-23
Release/1.3.1

hellais:
Opening this PR to hopefully aid the code review.


ooniprobe/Info.plist
```diff
@@ -11,11 +11,11 @@
 	<key>CFBundleInfoDictionaryVersion</key>
 	<string>6.0</string>
 	<key>CFBundleName</key>
-	<string>$(PRODUCT_NAME)</string>
```
hellais: Isn't it possible to just change the `PRODUCT_NAME` variable, instead of hardcoding `OONI Probe` in here?

ooniprobe/Model/NetworkMeasurement.mm
```diff
@@ -509,24 +540,28 @@ -(void)on_entry:(const char*)str{
...
         for (NSString *key in keys) {
-            if ([json objectForKey:key]){
-                if ([json objectForKey:key] == [NSNull null]) {
+            if ([[json objectForKey:@"test_keys"] objectForKey:key]){
+                if ([[json objectForKey:@"test_keys"] objectForKey:key] == [NSNull null]) {
                     if (blocking < ANOMALY_ORANGE)
                         blocking = ANOMALY_ORANGE;
                 }
-                else if ([[json objectForKey:key] isEqualToString:@"blocked"]) {
+                else if ([[[json objectForKey:@"test_keys"] objectForKey:key] isEqualToString:@"blocked"]) {
```
hellais: What happens if the value of the key is a boolean (which it can be)? Does calling `isEqualToString` cause some unhandled exception? (apologies for the noob obj-c question).

ooniprobe/Model/NetworkMeasurement.mm
```diff
@@ -652,22 +690,27 @@ -(void)on_entry:(const char*)str{
...
+                else if ([[[json objectForKey:@"test_keys"] objectForKey:key] boolValue]) {
                     blocking = ANOMALY_RED;
                 }
             }
         }
-        if (blocking > self.anomaly){
-            self.anomaly = blocking;
-            [TestStorage set_anomaly:self.test_id :blocking];
-        }
+        [self updateAnomaly:blocking];
```
hellais: As I mentioned on slack.

I think this whole file can do with some easy refactoring, that would improve readability and maybe also prevent errors like this one from emerging in the future.

I will paste for posterity my suggestions in here:

There is a fair amount of duplication in the above code. I was suggesting to refactor the block like:

````
        int blocking = ANOMALY_GREEN;
        NSArray *keys = [[NSArray alloc] initWithObjects:@"facebook_tcp_blocking", @"facebook_dns_blocking", nil];
        for (NSString *key in keys) {
            if ([json objectForKey:key]){
                if ([json objectForKey:key] == [NSNull null]) {
                    if (blocking < ANOMALY_ORANGE)
                        blocking = ANOMALY_ORANGE;
                }
                else if ([[json objectForKey:key] boolValue]) {
                    blocking = ANOMALY_RED;
                }
            }
        }
```

Into some function that takes arguments like so:

```
func isOrangeAnomaly(testKeys, targetKeys)
```

There are also many more bits of code that can probably be easily refactored into a function and can reduce the amount of duplication.

ooniprobe/Utility/Tests.h
```diff
@@ -3,6 +3,8 @@
 
 @interface Tests : NSObject
 @property (strong, nonatomic) NSMutableArray *availableNetworkMeasurements;
+@property (strong, nonatomic) NSMutableDictionary *availableNetworkMeasurementsDic;
```
hellais: It's generally more common to refer to "dictionaries" as "dict" rather than "dic".

hellais: I also don't see where this dictionary is even being used. Searching in this PR I don't see it used anywhere.

ooniprobe/Info.plist
lorenzoPrimi: I should check if it's possible without renaming every single folder called 'ooniprobe' in the project. I found this answer on stackoverflow

ooniprobe/Model/NetworkMeasurement.mm
lorenzoPrimi: according to the specifications I got it can't be, only string or null.
Or was your question more generic about what can happen in that case?

ooniprobe/Model/NetworkMeasurement.mm
lorenzoPrimi: this only for IM tests?

ooniprobe/Utility/Tests.h
lorenzoPrimi: in fact is not. thanks for pointing it out

ooniprobe/Info.plist
lorenzoPrimi: Ok it's fine to change PRODUCT_NAME, however it does some changes in the project settings like
 -           BuildableName = "ooniprobe.app"
 +           BuildableName = "OONI Probe.app"

should I proceed?

hellais: Yes

ooniprobe/Model/NetworkMeasurement.mm
hellais: Ah yes, you are right. I was confusing this with some other keys. In any case what happens if it isn't a string? If an exception is triggered, I think it would be wise to in any case be defensive and assert the type.

ooniprobe/Model/NetworkMeasurement.mm
hellais: There are very similar blocks of code in the other tests as well. I would expect that with this refactoring you can reduce the amount of overall code in this file significantly.

ooniprobe/Model/NetworkMeasurement.mm
lorenzoPrimi: The crash happens only in the NULL case.
If it's a bool it will be treated like "1" string

ooniprobe/Model/NetworkMeasurement.mm
lorenzoPrimi: The fact is that this class is working well and will be totally deleted in the 2.0 version. 
Refactoring the code means that everything should be tested again and can produce bugs.
I'm against that.

hellais:
@lorenzoPrimi did you update to the latest https://github.com/citizenlab/test-lists/blob/master/lists/global.csv? Be sure to do that before the release.

ooniprobe/Resources/test-lists/global.csv
```diff
@@ -222,8 +222,9 @@ https://riseup.net,HOST,Hosting and Blogging Platforms,2014-04-15,citizenlab,Upd
 https://rsf.org/,NEWS,News Media,2014-04-15,citizenlab,Updated by OONI on 2017-02-14
 http://russia.tv,NEWS,News Media,2014-04-15,citizenlab,Updated by OONI on 2017-02-14
 http://sakhr.com,COMT,Communication Tools,2014-04-15,citizenlab,Updated by OONI on 2017-02-14
-https://sci-hub.ac/,FILE,File-sharing,2016-07-13,David Fifield,Updated by OONI on 2017-02-14
-http://sci-hub.org/,FILE,File-sharing,2016-07-13,David Fifield,Updated by OONI on 2017-02-14
+https://sci-hub.ac/,FILE,File-sharing,2016-07-13,David Fifield,Updated by OONI on 2017-02-14; blocked as a result of court order 2017-11-17 https://www.theregister.co.uk/2017/11/23/scihubs_become_inactive_following_court_order/
+https://sci-hub.bz/,FILE,File-sharing,2017-11-23,David Fifield,
+http://sci-hub.org/,FILE,File-sharing,2016-07-13,David Fifield,Taken down by court order November 2015 https://torrentfreak.com/court-orders-shutdown-of-libgen-bookfi-and-sci-hub-151102/; updated by OONI on 2017-02-14
```
hellais: For the future, remember to put in the commit message the hash of the tip you have updated it to, so that we can know what it's dated as.
I assume you updated to dc6e07b63baaaa5774bb8d04c074f48a96a89d81

PR #182, opened 2018-09-11
Update/measurement kit 0.9

lorenzoPrimi:


ooniprobe/TestClasses/MKNetworkTest.mm
```diff
@@ -3,132 +3,258 @@
...
+-(void)runTest{
+    self.backgroundTask = [[UIApplication sharedApplication] beginBackgroundTaskWithExpirationHandler:^{
+        [[UIApplication sharedApplication] endBackgroundTask:self.backgroundTask];
+        self.backgroundTask = UIBackgroundTaskInvalid;
+    }];
+    dispatch_async(
+                   dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
+                       NSLog(@"%@", [self.settings getSerializedSettings]);
+                       mk_unique_task taskp{mk_nettest_start([[self.settings getSerializedSettings] UTF8String])};
+                       while (!mk_task_is_done(taskp.get())) {
```
lorenzoPrimi: @bassosimone you might want to quickly check this piece (until row 154)

Podfile
```diff
@@ -2,9 +2,11 @@ platform :ios, '9.0'
 # ignore all warnings from all pods
 inhibit_all_warnings!
 target 'ooniprobe' do
-    pod 'measurement_kit',
-        :git => 'https://github.com/measurement-kit/measurement-kit.git',
-	:tag => 'v0.8.3'
+  pod 'measurement_kit',
+    :git => 'https://github.com/measurement-kit/ios-libs.git'
```
bassosimone: I should probably add a tag here, but for now I guess we can avoid it, right?

ooniprobe/TestClasses/MKNetworkTest.mm
bassosimone: sure!

ooniprobe/TestClasses/MKNetworkTest.mm
```diff
@@ -3,132 +3,257 @@
...
+                           NSDictionary *value = [evinfo objectForKey:@"value"];
+                           if (key == nil || value == nil) {
+                               return;
+                           }
+                           if ([key isEqualToString:@"status.started"]) {
+                               //[self updateProgressBar:0];
+                           }
+                           else if ([key isEqualToString:@"status.measurement_start"]) {
+                               NSNumber *idx = [value objectForKey:@"idx"];
+                               NSString *input = [value objectForKey:@"input"];
```
bassosimone: For robustness, it's better to _not_ assume that fields will be in the JSON in the future, even though I _should not_ remove them without telling you. So, I would use the following pattern for each branch:

```ObjectiveC
  if ([key isEqualToString:@"<event-name>") {
    NSSomeType *type = [value objectForKey:@"<field-name>"];
    NSSomeOtherType *otherType = [value objectForKey:@"<other-field-name>"];
    // ...
    if (type && otherType /* && ... */) {
      // Either use `type` and `otherType` and `...` directly or pass them to a method
    }
  }
```

In this specific case, hence:

```ObjectiveC
                           else if ([key isEqualToString:@"status.measurement_start"]) {
                             NSNumber *idx = [value objectForKey:@"idx"];
                             NSString *input = [value objectForKey:@"input"];
                             if (idx && input) {
                               Measurement *measurement = [self createMeasurementObject];
                               if ([input length] > 0){
                                 measurement.url_id = [Url getUrl:input];
                                 [measurement save];
                               }
                               [self.measurements setObject:measurement forKey:idx];
                             } else {
                               NSLog(@"Either idx or input is nil");
                               assert(false); // crash if it's a debug build 
                             }
                           }
```

Does this makes sense to you?

Podfile
lorenzoPrimi: for now we can avoit it, later better use tags

ooniprobe/TestClasses/MKNetworkTest.mm
lorenzoPrimi: added null checks

Podfile
bassosimone: agreed!

ooniprobe/TestClasses/MKNetworkTest.mm
bassosimone: Thanks!

PR #199, opened 2018-12-18
Changes to use mkall-ios's CocoaPod

bassosimone:
I have tested this integration using the simulator with iPhone XR and iOS 12.1.

This seems to work as expected.

Thus, I believe we can say we have a CocoaPod.

Xref: measurement-kit/measurement-kit#1691 measurement-kit/measurement-kit#1703


lorenzoPrimi:
Everything compiles and run totally fine.

Issue #200, opened 2018-12-18
Continuous integration is broken


bassosimone:
We should repair continuous integration. It's a bummer to have the red X for all commits.

hellais:
Travis is green again thanks to: https://github.com/ooni/probe-ios/pull/204.

bassosimone:
@hellais thanks for working on this! <3

PR #204, opened 2019-01-03
Bump xcode version

hellais:
This is my attempt at fixing: https://github.com/ooni/probe-ios/issues/200


hellais:
The CI is now fixed.

PR #217, opened 2019-01-11
Fastlane deliver

hellais:
This adds support for pushing metadata and screenshots automatically to fastlane.

It will also require some changes to the ooni/translations repo.

For the moment I am using this to automatically upload the screenshots and app store descriptions.

TODO:

* [ ] Make changes in ooni/translations to write descriptions to fastlane/metadata/*
* [ ] Come up with a strategy for committing or including metadata in this repo


hellais:
This superseeds #216

hellais:
Note: I haven't committed the metadata directory tree, because I am not yet sure if we should do that or if we should instead just generate that from `ooni/translations`.

Issue #219, opened 2019-01-14
TestResultsTableViewController.m crash


hellais:
This crash comes from fabric:
```
Fatal Exception: NSInvalidArgumentException
*** -[__NSDictionaryM setObject:forKey:]: key cannot be nil
```

stacktrace of the error is:

```
Fatal Exception: NSInvalidArgumentException
0  CoreFoundation                 0x1d622cec4 __exceptionPreprocess
1  libobjc.A.dylib                0x1d53fda40 objc_exception_throw
2  CoreFoundation                 0x1d61a4494 _CFArgv
3  CoreFoundation                 0x1d6117a0c -[__NSDictionaryM setObject:forKey:]
4  ooniprobe                      0x104be6164 -[TestResultsTableViewController testFilter:] (TestResultsTableViewController.m:39)
5  ooniprobe                      0x104bfa85c __42-[ResultsHeaderViewController reloadQuery]_block_invoke (ResultsHeaderViewController.m:50)
6  libdispatch.dylib              0x1d5c656c8 _dispatch_call_block_and_release
7  libdispatch.dylib              0x1d5c66484 _dispatch_client_callout
8  libdispatch.dylib              0x1d5c45b34 _dispatch_main_queue_callback_4CF$VARIANT$armv81
9  CoreFoundation                 0x1d61bcdf4 __CFRUNLOOP_IS_SERVICING_THE_MAIN_DISPATCH_QUEUE__
10 CoreFoundation                 0x1d61b7cbc __CFRunLoopRun
11 CoreFoundation                 0x1d61b71f0 CFRunLoopRunSpecific
12 GraphicsServices               0x1d8430584 GSEventRunModal
13 UIKitCore                      0x20350ed40 UIApplicationMain
14 ooniprobe                      0x104beefc8 main (main.m:6)
15 libdyld.dylib                  0x1d5c76bb4 start
```

bassosimone:
@hellais what is the Fabric ID of this crash? (I _think_ it may help me to know that.)

bassosimone:
I think I figured it out, it's `#66` on Fabric.

bassosimone:
Now I'm trying to understand what error condition is this.

lorenzoPrimi:
You are right, either in this case `current.start_time` is `nil` but this is really unlikely 
OR `stringFromDate` returns a nil object.
I will investigate more

lorenzoPrimi:
The start_time can't be null as it's initialized when the db object is initialized:
[...]

PR #236, opened 2019-02-06
Dev

lorenzoPrimi:
## Changelog
* Updated mkall to 0.3.10
* Fixes https://github.com/ooni/probe-ios/issues/219
* Various UI small fixes

### DIFF
--- a/ooniprobe/Model/Database/Measurement.mm
+++ b/ooniprobe/Model/Database/Measurement.mm
@@ -51,9 +51,6 @@
 }
 
 -(NSString*)getLocalizedStartTime{
-    //TODO we might need a copy for this
-    if (self.start_time == nil)
-        return NSLocalizedString(@"TestResults.UnknownASN", nil);
     //from https://developer.apple.com/library/content/documentation/MacOSX/Conceptual/BPInternational/InternationalizingLocaleData/InternationalizingLocaleData.html
     NSString *localizedDateTime = [NSDateFormatter localizedStringFromDate:self.start_time dateStyle:NSDateFormatterShortStyle timeStyle:NSDateFormatterShortStyle];
     return localizedDateTime;
--- a/ooniprobe/Model/Database/Measurement.mm
+++ b/ooniprobe/Model/Database/Measurement.mm
@@ -50,6 +50,15 @@
     return [self.result_id getLogFile:self.test_name];
 }
 
+-(NSString*)getLocalizedStartTime{
+    //TODO we might need a copy for this
+    if (self.start_time == nil)
+        return NSLocalizedString(@"TestResults.UnknownASN", nil);
+    //from https://developer.apple.com/library/content/documentation/MacOSX/Conceptual/BPInternational/InternationalizingLocaleData/InternationalizingLocaleData.html
+    NSString *localizedDateTime = [NSDateFormatter localizedStringFromDate:self.start_time dateStyle:NSDateFormatterShortStyle timeStyle:NSDateFormatterShortStyle];
+    return localizedDateTime;
+}
+
 -(void)save{
     [self commit];
     /*
--- a/ooniprobe/Model/Database/Result.mm
+++ b/ooniprobe/Model/Database/Result.mm
@@ -111,6 +111,14 @@
     return [NSString stringWithFormat:@"%@-%@.log", self.Id, test_name];
 }
 
+-(NSString*)getLocalizedStartTime{
+    //TODO we might need a copy for this
+    if (self.start_time == nil)
+        return NSLocalizedString(@"TestResults.UnknownASN", nil);
+    NSString *localizedDateTime = [NSDateFormatter localizedStringFromDate:self.start_time dateStyle:NSDateFormatterShortStyle timeStyle:NSDateFormatterShortStyle];
+    return localizedDateTime;
+}
+
 -(void)save{
     [self commit];
     /*
--- a/ooniprobe/Model/Database/Result.mm
+++ b/ooniprobe/Model/Database/Result.mm
@@ -112,9 +112,6 @@
 }
 
 -(NSString*)getLocalizedStartTime{
-    //TODO we might need a copy for this
-    if (self.start_time == nil)
-        return NSLocalizedString(@"TestResults.UnknownASN", nil);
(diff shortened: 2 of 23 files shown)


ooniprobe/Model/Database/Measurement.mm
```diff
@@ -50,6 +50,15 @@ -(NSString*)getLogFile{
     return [self.result_id getLogFile:self.test_name];
 }
 
+-(NSString*)getLocalizedStartTime{
+    //TODO we might need a copy for this
+    if (self.start_time == nil)
+        return NSLocalizedString(@"TestResults.UnknownASN", nil);
```
hellais: Should this not be a different message string?

ooniprobe/Model/Database/Measurement.mm
```diff
@@ -50,6 +50,15 @@ -(NSString*)getLogFile{
...
+    //from https://developer.apple.com/library/content/documentation/MacOSX/Conceptual/BPInternational/InternationalizingLocaleData/InternationalizingLocaleData.html
+    NSString *localizedDateTime = [NSDateFormatter localizedStringFromDate:self.start_time dateStyle:NSDateFormatterShortStyle timeStyle:NSDateFormatterShortStyle];
```
hellais: What happens if `self.start_time` is not a valid date (ex. it is the string `I am an invalid date`)?

ooniprobe/Model/Database/Measurement.mm
lorenzoPrimi: No, as I decided with @agrabeli, as this is a temp fix we use this string to don't wait for translations. This fix is not present in Android and will be removed in iOS sometime in the future

ooniprobe/Model/Database/Result.mm
```diff
@@ -111,6 +111,14 @@ -(NSString*)getLogFile:(NSString*)test_name{
     return [NSString stringWithFormat:@"%@-%@.log", self.Id, test_name];
 }
 
+-(NSString*)getLocalizedStartTime{
+    //TODO we might need a copy for this
+    if (self.start_time == nil)
+        return NSLocalizedString(@"TestResults.UnknownASN", nil);
+    NSString *localizedDateTime = [NSDateFormatter localizedStringFromDate:self.start_time dateStyle:NSDateFormatterShortStyle timeStyle:NSDateFormatterShortStyle];
+    return localizedDateTime;
```
hellais: This seems to be the same code as the above. Can we refactor this into a common function instead of it being a class method?

ooniprobe/Model/Database/Measurement.mm
lorenzoPrimi: `start_time` can't be an invalid date.

ooniprobe/View/TestResults/TestResultsTableViewController.m
```diff
@@ -70,6 +70,8 @@ - (NSAttributedString *)titleForEmptyDataSet:(UIScrollView *)scrollView
 - (NSString *)tableView:(UITableView *)tableView titleForHeaderInSection:(NSInteger)section
 {
     NSString *month = [keys objectAtIndex:section];
+    if ([month isEqualToString:@"0000-00"])
+        return [NSLocalizedString(@"TestResults.UnknownASN", nil) uppercaseString];
```
hellais: Probably this NSLocalzedString also need to be a different value.

ooniprobe/Test/Test/AbstractTest.mm
```diff
@@ -275,8 +275,12 @@ -(void)writeString:(NSString*)str toFile:(NSString*)fileName{
 }
 
 -(void)onEntry:(JsonResult*)json obj:(Measurement*)measurement{
-    [self.result setStart_time:json.test_start_time];
-    [measurement setStart_time:json.test_start_time];
+    if (json.test_start_time != nil){
```
lorenzoPrimi: @hellais this is the reason why `start_time` is not an invalid date

ooniprobe/Model/Database/Measurement.mm
hellais: > `start_time` can't be an invalid date.

Where do you get this guarantee from? My understanding is that it is a string type so I believe it can be anything.

ooniprobe/Test/Test/AbstractTest.mm
hellais: I still don't see where you check the format of the string :|

ooniprobe/Model/Database/Measurement.mm
lorenzoPrimi: it's an nsdate not a string

ooniprobe/Test/Test/AbstractTest.mm
lorenzoPrimi: Again, it's not a string, it's a NSDate

ooniprobe/Model/Database/Measurement.mm
hellais: Ok I see. Maybe `TestResults.NotAvailable` is more appropriate though.

ooniprobe/Model/Database/Measurement.mm
hellais: Ah ok I understand now. Thanks for clarifying this.

ooniprobe/Test/Test/AbstractTest.mm
hellais: Where does the conversion from string inito NSDate happen?

lorenzoPrimi: line 242

ooniprobe/View/TestResults/TestResultsTableViewController.m
```diff
@@ -33,15 +33,17 @@ -(void)testFilter:(SRKQuery*)newQuery{
...
         NSMutableArray *arr = [[NSMutableArray alloc] init];
         NSString *key = [df stringFromDate:current.start_time];
-        if ([dic objectForKey:key])
-            arr = [[dic objectForKey:key] mutableCopy];
-        [arr addObject:current];
         if (key == nil){
-            //log object
+            //reporting error to fabric and delete test
             [CrashlyticsKit recordError:[NSError errorWithDomain:@"key_nil" code:0 userInfo:[current dictionary]]];
+            [current deleteObject];
```
lorenzoPrimi: @hellais better this way?

PR #238, opened 2019-02-12
Feature/send email

lorenzoPrimi:
Fixes https://github.com/ooni/probe/issues/819


ooniprobe/View/Settings/SettingsCategoriesTableViewController.m
```diff
@@ -46,11 +48,38 @@ - (void)tableView:(UITableView *)tableView didSelectRowAtIndexPath:(NSIndexPath
...
 
+-(void)sendEmail{
+    if ([MFMailComposeViewController canSendMail]){
+        MFMailComposeViewController *mailer = [[MFMailComposeViewController alloc] init];
+        mailer.mailComposeDelegate = self;
+        [mailer setSubject:[NSString stringWithFormat:@"[bug-report] OONI Probe iOS %@", [[[NSBundle mainBundle] infoDictionary] valueForKey:@"CFBundleShortVersionString"]]];
+        NSArray *toRecipients = [NSArray arrayWithObjects:@"contact@openobservatory.org", nil];
+        [mailer setToRecipients:toRecipients];
+        NSString *device = [NSString stringWithFormat:@"</br></br></br>app_version %@</br>device_model %@</br>os_version %@", [[[NSBundle mainBundle] infoDictionary] valueForKey:@"CFBundleShortVersionString"], [UIDevice currentDevice].modelName, [UIDevice currentDevice].systemVersion];
+        [mailer setMessageBody:device isHTML:YES];
```
hellais: Can we make the email that users send not HTML by default? I think this will make it easier for us to process them in the future if we want to.

ooniprobe/View/Settings/SettingsCategoriesTableViewController.m
```diff
@@ -46,11 +48,38 @@ - (void)tableView:(UITableView *)tableView didSelectRowAtIndexPath:(NSIndexPath
...
         [self performSegueWithIdentifier:@"toSettings" sender:self];
     [tableView deselectRowAtIndexPath:indexPath animated:YES];
 }
 
+-(void)sendEmail{
+    if ([MFMailComposeViewController canSendMail]){
+        MFMailComposeViewController *mailer = [[MFMailComposeViewController alloc] init];
+        mailer.mailComposeDelegate = self;
+        [mailer setSubject:[NSString stringWithFormat:@"[bug-report] OONI Probe iOS %@", [[[NSBundle mainBundle] infoDictionary] valueForKey:@"CFBundleShortVersionString"]]];
+        NSArray *toRecipients = [NSArray arrayWithObjects:@"contact@openobservatory.org", nil];
```
hellais: We should set the to field to be the `bugs@openobservatory.org` alias

NOTICE
```diff
@@ -40,3 +40,7 @@ Terms: https://github.com/MatthewYork/DateTools/blob/master/LICENSE
 - OCMapper 
 Url: https://github.com/aryaxt/OCMapper
 Terms: https://github.com/aryaxt/OCMapper/blob/master/License.txt
+
+- UIDevice-Hardware
+Url: https://github.com/erichoracek/UIDevice-Hardware
+Terms: https://github.com/erichoracek/UIDevice-Hardware/blob/master/LICENSE
```
hellais: Are we sure we need to use this third party library?

It seems like all it does is map `getSysInfoByName:"hw.machine"];` to a slightly more user-friendly string: https://github.com/erichoracek/UIDevice-Hardware/blob/master/UIDevice-Hardware.m#L44.

Since the model identifier needs to be used by us, we can look it up ourselves and not rely on third party code.

lorenzoPrimi: Well iPhone X is much more readable than iPhone 3,4
I can even copy this function inside the app and keep it updated.

ooniprobe/View/Settings/SettingsCategoriesTableViewController.m
lorenzoPrimi: So update the ticket. In the instruction we have to use "contact@openobservatory.org"

ooniprobe/View/Settings/SettingsCategoriesTableViewController.m
lorenzoPrimi: ok

ooniprobe/View/Settings/SettingsCategoriesTableViewController.m
hellais: Updated. I had listed the new alias in a comment, but it got a bit buried.

NOTICE
hellais: I think that since this info is developer information for us to investigate the bug, we can do the lookup ourselves.

If you still think it's useful to you to have it be resolved on the users device, then maybe vendorising (copy that bit of code into our codebase) just that function is the way to go as it's pretty small.
Keep in mind, though, that we will have to update this function every time a new phone comes out and it's probably easier to just use the model identifier and maintain a lookup table that is external to the apps.

I am thinking of this from the perspective of maybe integrating the bug reporting email alias inside of some automated bug reporting workflow.

lorenzoPrimi: I'm still not sold on this. 
This library is always updated (at least until now) and even if it will not be we will get the iPhone 15,75 string. (and in that case we can clone it and mantain ourselves)
It's much more difficult for us to "translate" the iPhone version to the model name every time we get an email.

hellais: Is there a reason why you need to have the brand name of the phone and the model ID is not enough?

My concern is that you may end up with two users (which have two different version of the app with two different version of this library) submitting a different model string, but actually are on the same phone.

For example say the `iPhone XI` comes out (`iPhone12,8`) and it takes us some time to ship an update to the library, we may have the same time people sending us `iPhone XI` & `iPhone 12,8`, which could be confusing.

In any case I don't think it's a super big deal, but if we can avoid including a third party library (which is more code that can include more bugs) it would be better imho.

lorenzoPrimi: It is enough but You will have to go trought every model manually and "translate" it into human readable string. 
Even if we have double string `iPhone XI` and `iPhone12,8` for a short period of time I don't think is such a big deal.
As you are the master of automation, automatizing every single bit, I don't really know why you prefer this solution!

ooniprobe/View/Settings/SettingsCategoriesTableViewController.m
lorenzoPrimi: Updated

ooniprobe/View/Settings/SettingsCategoriesTableViewController.m
lorenzoPrimi: Updated

NOTICE
lorenzoPrimi: @hellais your call 
I still prefer automating this little bit of translating model into names, but it's up to you.

lorenzoPrimi:
This PR is clearly blocked by the library/nonlibrary decision.
Even if it's totally bikeshedding I think would be a good test for similar (and maybe bigger) situations, when we reach a deadlock in decision.
@gabelula what is your suggestion to do in similar cases?
Should we ask different opinions of dev such as @bassosimone ?

bassosimone:
@lorenzoPrimi I believe it's correct to ask to a third person. Here's my view: we should collect raw data in the app and use this library on the server/support side when we receive a bug. We should not add more rope around our neck, there is already too much.

lorenzoPrimi:
@hellais feel free to review it again. I removed the library and copied the needed method into the app

ooniprobe/Utility/UIDevice-Hardware.m
```diff
@@ -0,0 +1,31 @@
...
+    free(answer);
+    return results;
+}
+
+- (NSString *)modelIdentifier
+{
+    return [self getSysInfoByName:"hw.machine"];
+}
+
+@end
```
hellais: Do we still need this code since we are dropping the dependency?

Is it maybe not possible to call the `getSysInfoByName` method in the function creating the email directly?

hellais: This comment may also due to my lack of understanding how objective-c patterns work, but perhaps it's sufficient to just define a function called `getSysInfoByName` inside of `SettingsCategoriesTableViewController.m`.

lorenzoPrimi: Yes we still need this code as we need to extend the UIDevice class to be able to call that function.
Also `getSysInfoByName` doesn't belong to a ViewController, let's keep the code clean and in order.

lorenzoPrimi: this is one of the reason why I would have kept the library in the first place, one less class to worry about in the main project

hellais: Ok understood. Then let's perhaps call it a name different than `UIDevice-Hardware` so that we avoid confusing this with the library we have dropped called `UIDevice-Hardware` (https://github.com/erichoracek/UIDevice-Hardware)?

ooniprobe/Utility/UIDevice-Hardware.m
```diff
@@ -0,0 +1,31 @@
...
+
+@implementation UIDevice (Hardware)
+
+- (NSString *)getSysInfoByName:(char *)typeSpecifier
+{
+    size_t size;
+    sysctlbyname(typeSpecifier, NULL, &size, NULL, 0);
+
+    char *answer = malloc(size);
```
hellais: @bassosimone probably knows this better, but I believe you need to check if the return value of malloc is null to verify if you have allocated memory successfully.

ooniprobe/Utility/UIDevice-Hardware.m
lorenzoPrimi: Any idea for a new name? UIDevice-Extended?

ooniprobe/Utility/UIDevice-Hardware.m
lorenzoPrimi: I know nothing about this code. It's the same code of the library, of iOS cookbook and of many github answers so I guess it's the way to go 🤷‍♂️

ooniprobe/Utility/UIDevice-Hardware.m
```diff
@@ -0,0 +1,31 @@
...
+
+@interface UIDevice (Hardward)
+
+- (NSString *)modelNameForModelIdentifier:(NSString *)modelIdentifier;
+
+@end
+
+@implementation UIDevice (Hardware)
+
+- (NSString *)getSysInfoByName:(char *)typeSpecifier
```
hellais: Since this method is only ever called with `typeSpecifier` set to `hw.machine` maybe we can just place all this code inside of the `modelIdentifier` method and explicit the `hw.machine` string to avoid an extra un-necessary layer of indirection.

lorenzoPrimi: Can we avoid bikeshed too much for such a simple function? We are talking about sending an email to support 😅 I just want to merge this and move on with more important tasks

ooniprobe/Utility/UIDevice-Hardware.m
hellais: `DeviceInfo`?

lorenzoPrimi: Good call

ooniprobe/Utility/UIDevice-Hardware.m
bassosimone: So, it really depends on the VMM. Generic code _should_ check the return value of `malloc` and call `abort` if it's `NULL`. On Linux, and possibly also on macOS, the VMM does not fail in its default configuration and later, if there's not enough core, you get a segfault. My suggestion is to check the return value and `abort` if it's `NULL`.

ooniprobe.xcodeproj/project.pbxproj
```diff
@@ -1366,7 +1372,7 @@
 				STRIP_INSTALLED_PRODUCT = NO;
 				TARGETED_DEVICE_FAMILY = "1,2";
 				USER_HEADER_SEARCH_PATHS = "";
-				VALID_ARCHS = "armv64 armv7s i386 x86_64";
+				VALID_ARCHS = "arm64 armv7s i386 x86_64";
```
bassosimone: How come that this change _also_ appears here? Didn't we fix this long time ago?

ooniprobe/AppDelegate.mm
```diff
@@ -24,8 +24,9 @@ - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(
     [[UIBarButtonItem appearance] setTitleTextAttributes:[NSDictionary dictionaryWithObjectsAndKeys: [UIFont fontWithName:@"FiraSans-Regular" size:16],NSFontAttributeName, nil] forState:UIControlStateNormal];
     [[UINavigationBar appearance] setBarTintColor:[UIColor colorWithRGBHexString:color_blue5 alpha:1.0f]];
     [[UINavigationBar appearance] setTranslucent:FALSE];
-    
-    #ifdef RELEASE
+    [[UINavigationBar appearance] setTintColor:[UIColor colorWithRGBHexString:color_white alpha:1.0f]];
+
+  #ifdef RELEASE
```
bassosimone: Please, keep the same indent as before.

bassosimone: (I mean, I don't like this way of indenting `#ifdef`s, but at least keep it consistent please!)

ooniprobe/Images.xcassets/settings/about_ooni.imageset/Contents.json
```diff
@@ -8,5 +8,8 @@
   "info" : {
     "version" : 1,
     "author" : "xcode"
+  },
+  "properties" : {
+    "template-rendering-intent" : "template"
```
bassosimone: Why is this change appearing here?

ooniprobe/Utility/UIDevice-Hardware.m
lorenzoPrimi: https://books.google.ru/books?id=0DdbblZ6NlYC&pg=PT37&lpg=PT37&dq=%22hw.machine%22+cookbook&source=bl&ots=oTYPC7igAQ&sig=ACfU3U3nZWK4v4ZcvbSkKuo-BstTwPA9FQ&hl=it&sa=X&ved=2ahUKEwjHu-Os2I7hAhWv-ioKHf7oAe8Q6AEwAHoECAEQAQ#v=onepage&q=%22hw.machine%22%20cookbook&f=false

ooniprobe/Utility/UIDevice-DeviceInfo.m
```diff
@@ -0,0 +1,25 @@
+#include <sys/sysctl.h>
+#import "UIDevice-DeviceInfo.h"
+
+@implementation UIDevice (DeviceInfo)
+
+- (NSString *)getSysInfoByName:(char *)typeSpecifier
+{
+    size_t size;
+    sysctlbyname(typeSpecifier, NULL, &size, NULL, 0);
```
bassosimone: `sysctlbyname` returns `0` on success and nonzero on failure. Please check the return value and `abort` if it returns a failure. I assume this kind of functions should not fail until we get the params right.

ooniprobe/Utility/UIDevice-DeviceInfo.m
```diff
@@ -0,0 +1,25 @@
+#include <sys/sysctl.h>
+#import "UIDevice-DeviceInfo.h"
+
+@implementation UIDevice (DeviceInfo)
+
+- (NSString *)getSysInfoByName:(char *)typeSpecifier
```
bassosimone: Please make the function argument a `const char *`.

ooniprobe/Utility/UIDevice-DeviceInfo.m
```diff
@@ -0,0 +1,25 @@
...
+{
+    size_t size;
```
bassosimone: Please, initialize `size` to zero.

ooniprobe/Utility/UIDevice-DeviceInfo.m
bassosimone: Please, also `abort` if the returned value is `0` and `size` is `0`

ooniprobe/Utility/UIDevice-DeviceInfo.m
```diff
@@ -0,0 +1,25 @@
...
+    sysctlbyname(typeSpecifier, NULL, &size, NULL, 0);
+
+    char *answer = malloc(size);
```
bassosimone: I could not find in macOS manual pages whether the terminating `\0` is included. According to OpenBSD manual pages it is, but that does not provide us the guarantee that the same happens for Darwin.

ooniprobe/Utility/UIDevice-Hardware.m
bassosimone: Not checking the return value of `malloc` is not a good programming practice. It leads to headaches when porting code to other systems. High performance code may decide not to do that if the platform where this happens is clearly one where `malloc` does not return `NULL` in case of failure. This is not the case. Please, check the return value of `malloc`.

ooniprobe/Utility/UIDevice-Hardware.m
bassosimone: > I just want to merge this and move on with more important tasks

I disagree with this view. While I understand why you want to do more challenging stuff, and I really sympathise with this feeling, it's also important that we try to minimise the amount of unnecessary code that we commit, in the interest of maintainability.

ooniprobe.xcodeproj/project.pbxproj
lorenzoPrimi: yes we did, but this branch is oooold

ooniprobe/Utility/UIDevice-DeviceInfo.m
lorenzoPrimi: added other exception, I don't know how to call it

ooniprobe.xcodeproj/project.pbxproj
bassosimone: There should be a GitHub button offering you to merge master into this branch. This should be such that see a clean diff between this branch and master. Unless @hellais has objections, I really suggest to use it.

ooniprobe/Utility/UIDevice-DeviceInfo.m
bassosimone: `parse_error`

bassosimone: Ah, I see what you mean. You're using ObjectiveC exceptions.

To call `abort` just do `abort();`

lorenzoPrimi: So should I use abort or Objc exceptions? what is suggested?

ooniprobe.xcodeproj/project.pbxproj
lorenzoPrimi: master already has more PRs merged in that, is still possible to do what you say without breaking everything?

lorenzoPrimi: After the rebase this is still different. What is correct the one in this branch or the one in master?

ooniprobe/Utility/UIDevice-Hardware.m
bassosimone: Can we resolve this convo, @lorenzoPrimi ?

ooniprobe.xcodeproj/project.pbxproj
bassosimone: Why did you rebase?

ooniprobe/Utility/UIDevice-DeviceInfo.m
bassosimone: I'm a bit out of my depth here. My understanding is that `abort` should not call `malloc`. I do not know what happens when raising an ObjC exception. My sense is that, if `malloc` fails by returning `NULL`, then the only sane thing is to crash. So, even without knowing about the ObjC exception mechanism details, I'd say it's fine to call `abort` here. (I also checked whether `_exit` would be a better alternative, but the Darwin manual page mentions too many things that `_exit` does, compared to `abort`, so I'm leaning towards `abort` here.)

ooniprobe/Images.xcassets/settings/about_ooni.imageset/Contents.json
bassosimone: So, wait, why did you resolve the conversation without responding?

ooniprobe/Utility/UIDevice-DeviceInfo.m
```diff
@@ -0,0 +1,40 @@
...
+#import "UIDevice-DeviceInfo.h"
+
+@implementation UIDevice (DeviceInfo)
+
+- (NSString *)getSysInfoByName:(const char *)typeSpecifier
+{
+    size_t size = 0;
+    int retsize = sysctlbyname(typeSpecifier, NULL, &size, NULL, 0);
+
+    if (retsize == 0 || size == 0){
```
bassosimone: Please, call this variable just `ret`

ooniprobe/Utility/UIDevice-DeviceInfo.m
```diff
@@ -0,0 +1,40 @@
...
+
+    char *answer = malloc(size);
+    if (answer == NULL) {
+        NSException* myException = [NSException
+                                    exceptionWithName:@"MallocError"
+                                    reason:@"Malloc Error"
+                                    userInfo:nil];
+        @throw myException;
+    }
+    sysctlbyname(typeSpecifier, answer, &size, NULL, 0);
```
bassosimone: You also need to check the return value here

ooniprobe/Utility/UIDevice-Hardware.m
lorenzoPrimi: Sure, it has been renamed

ooniprobe.xcodeproj/project.pbxproj
lorenzoPrimi: I didn't rebase, I updated from master. Wrong choice of terms

ooniprobe/Images.xcassets/settings/about_ooni.imageset/Contents.json
lorenzoPrimi: Maybe I did by mistake. This is a change that should be here. it's an image that changed

ooniprobe/Utility/UIDevice-Hardware.m
lorenzoPrimi: As the previous function became really long I would still keep this one, so we can reuse it to get more insight from the hardware (maybe in the future) by passing different params.

ooniprobe/Utility/UIDevice-DeviceInfo.m
lorenzoPrimi: Is there any change to do here?

ooniprobe/Utility/UIDevice-DeviceInfo.m
lorenzoPrimi: how to call the return value?

ooniprobe/Images.xcassets/settings/about_ooni.imageset/Contents.json
lorenzoPrimi: After checking another branch I noticed is an useless change as this image is used in setting and it has already colors.

ooniprobe/Utility/UIDevice-Hardware.m
bassosimone: > As the previous function became really long

I would not call it really long. But I sympathize with the feeling that short functions are good.

> so we can reuse it to get more insight from the hardware

Yeah, this is also a good argument. I'm going to resolve this convo.

ooniprobe.xcodeproj/project.pbxproj
bassosimone: Ah, okay. Well, let me double check what is going on here.

ooniprobe/Images.xcassets/settings/about_ooni.imageset/Contents.json
bassosimone: Ok, thanks for clarifying.

ooniprobe/Utility/UIDevice-DeviceInfo.m
bassosimone: I think I should do some further research

ooniprobe/Images.xcassets/settings/about_ooni.imageset/Contents.json
lorenzoPrimi: My bad, this is needed to keep that image attributes in "sync" with others that are all considered "template".
Template is an image that can be colored used setTintColor

ooniprobe.xcodeproj/project.pbxproj
lorenzoPrimi: `arm64` should be the good one and `armv64` the wrong one, so this should be good to merge

ooniprobe/Utility/UIDevice-DeviceInfo.m
lorenzoPrimi: can we resolve this?

ooniprobe.xcodeproj/project.pbxproj
bassosimone: @lorenzoPrimi I am starting to be quite confused at this point:

```
$ git branch
  feature/send_email
  fix/crashes
* master
$ git describe --tags
v2.0.5-22-g3cb279e7
$ git grep armv64
ooniprobe.xcodeproj/project.pbxproj:                            VALID_ARCHS = "armv64 armv7s i386 x86_64";
[0]dunkerque:probe-ios$ 
```

Am I using the wrong branch? I would have expected the issue to be fixed in master? Am I doing other doofus?

ooniprobe/Utility/UIDevice-DeviceInfo.h
```diff
@@ -0,0 +1,10 @@
+#import <UIKit/UIKit.h>
```
bassosimone: I'd like to learn something today. Why is the file named `UIDevice-DeviceInfo`? Is this because UIDevice implements the DeviceInfo protocol? (Does this make sense?)

ooniprobe/Utility/UIDevice-DeviceInfo.m
```diff
@@ -0,0 +1,29 @@
+#include <sys/sysctl.h>
+#import "UIDevice-DeviceInfo.h"
+
+@implementation UIDevice (DeviceInfo)
+
+- (NSString *)getSysInfoByName:(const char *)typeSpecifier
+{
+    size_t size = 0;
+    int ret = sysctlbyname(typeSpecifier, NULL, &size, NULL, 0);
+    if (ret == 0 || size == 0) abort();
```
bassosimone: Did you test this code?

ooniprobe/Utility/UIDevice-DeviceInfo.m
```diff
@@ -0,0 +1,29 @@
...
+
+    char *answer = malloc(size);
+    if (answer == NULL) abort();
+    
+    int ret2 = sysctlbyname(typeSpecifier, answer, &size, NULL, 0);
```
bassosimone: Just reuse `ret` from above.

```objc
    ret2 = sysctlbyname(...);
```

ooniprobe/Utility/UIDevice-DeviceInfo.m
```diff
@@ -0,0 +1,29 @@
...
+    if (ret2 == 0) abort();
```
bassosimone: Did you test this code?

bassosimone: Here you're leaking `answer`. Please refactor to become:

```C
    if (ret == 0) {
        free(answer);
        abort();
    }
```

It's true that after `abort` you don't need to worry about `malloc`-ed memory. Yet, I'd like to see correct code that is more likely to remain correct after some refactoring.

ooniprobe.xcodeproj/project.pbxproj
lorenzoPrimi: I dont remember when it was fixed. Probably XCode override something?

ooniprobe/Utility/UIDevice-DeviceInfo.h
lorenzoPrimi: Because it's a subclass of UIDevice and @hellais didn't want to call it UIDevice-Hardware as the library and he proposed DeviceInfo :D

ooniprobe/Utility/UIDevice-DeviceInfo.m
lorenzoPrimi: Just tested on the simulator. ret is 0 and crashes the code.

lorenzoPrimi: maybe we should check the ret value ?

lorenzoPrimi: removing both if (ret ==0) works

ooniprobe.xcodeproj/project.pbxproj
bassosimone: I don't know. But we should have a better process and not loose fixes. I guess the workflow with master and possibly many small PRs to master and code review is good to prevent that.

ooniprobe/Utility/UIDevice-DeviceInfo.h
bassosimone: Ok thanks for the context. I agree to name it differently FWIW.

ooniprobe/Utility/UIDevice-DeviceInfo.m
bassosimone: `sysctlbyname` returns `0` on success so you should write `ret != 0`

ooniprobe/View/Settings/SettingsCategoriesTableViewController.m
```diff
@@ -46,11 +48,38 @@ - (void)tableView:(UITableView *)tableView didSelectRowAtIndexPath:(NSIndexPath
...
+        mailer.mailComposeDelegate = self;
+        [mailer setSubject:[NSString stringWithFormat:@"[bug-report] OONI Probe iOS %@", [[[NSBundle mainBundle] infoDictionary] valueForKey:@"CFBundleShortVersionString"]]];
+        NSArray *toRecipients = [NSArray arrayWithObjects:@"bugs@openobservatory.org", nil];
+        [mailer setToRecipients:toRecipients];
+        NSString *device = [NSString stringWithFormat:@"\n\n\napp_version %@\ndevice_model %@\nos_version %@", [[[NSBundle mainBundle] infoDictionary] valueForKey:@"CFBundleShortVersionString"], [UIDevice currentDevice].modelIdentifier, [UIDevice currentDevice].systemVersion];
+        [mailer setMessageBody:device isHTML:NO];
+        [[mailer navigationBar] setTranslucent:NO];
+        [[mailer navigationBar] setTintColor:[UIColor colorWithRGBHexString:color_white alpha:1.0f]];
+        [[mailer navigationBar] setTitleTextAttributes: [NSDictionary dictionaryWithObjectsAndKeys: [UIColor colorWithRGBHexString:color_white alpha:1.0f], NSForegroundColorAttributeName, [UIFont fontWithName:@"FiraSans-SemiBold" size:16], NSFontAttributeName, nil]];
+        [self presentViewController:mailer animated:YES completion:nil];
```
bassosimone: Where are you using the new class?

ooniprobe/Utility/UIDevice-DeviceInfo.m
bassosimone: Yes, let's use `abort`.

ooniprobe/View/Settings/SettingsCategoriesTableViewController.m
lorenzoPrimi: [UIDevice currentDevice].modelIdentifier

ooniprobe/Utility/UIDevice-DeviceInfo.m
lorenzoPrimi: done

ooniprobe/Utility/UIDevice-DeviceInfo.m
lorenzoPrimi: oh I see. tested and working

ooniprobe/View/Settings/SettingsCategoriesTableViewController.m
bassosimone: What I don't get is why you're using the parent class and not the derived class. I guess this has to do with my little understanding of ObjectiveC, so an explanation will help me to know this language better and be more useful in the future.

lorenzoPrimi: The new class has all the method of the basic class+ the new ones.
By including the new class in the header all the methods to UIDevice will be called on the new class. Don't really know how but it's how it works

Issue #222, opened 2019-01-17
Crash ObjectMapper.m EXC_BAD_ACCESS KERN_INVALID_ADDRESS


hellais:
This has affected 26 users and is `#56` on fabric.

```
Crashed: com.apple.root.default-qos
0  libobjc.A.dylib                0x1b18b0d70 objc_msgSend + 16
1  CoreFoundation                 0x1b25b9b08 -[__NSDictionaryM setObject:forKeyedSubscript:] + 524
2  ooniprobe                      0x100e0681c -[ObjectMapper typeForProperty:andClass:] (ObjectMapper.m:545)
3  ooniprobe                      0x100e047e4 -[ObjectMapper processDictionaryFromObject:] (ObjectMapper.m:163)
4  ooniprobe                      0x100e042bc -[ObjectMapper dictionaryFromObject:] (ObjectMapper.m:113)
5  ooniprobe                      0x100e04a74 -[ObjectMapper processDictionaryFromObject:] (ObjectMapper.m:176)
6  ooniprobe                      0x100e042bc -[ObjectMapper dictionaryFromObject:] (ObjectMapper.m:113)
7  ooniprobe                      0x100e03a6c -[NSObject(ObjectMapper) dictionary] (NSObject+ObjectMapper.m:39)
8  ooniprobe                      0x100daa704 -[Settings getSettingsDictionary] (Settings.m:23)
9  ooniprobe                      0x100dc2424 __24-[MKNetworkTest runTest]_block_invoke.36 (MKNetworkTest.mm:40)
10 libdispatch.dylib              0x1b21016c8 _dispatch_call_block_and_release + 24
11 libdispatch.dylib              0x1b2102484 _dispatch_client_callout + 16
12 libdispatch.dylib              0x1b20a5490 _dispatch_queue_override_invoke + 680
13 libdispatch.dylib              0x1b20b1a18 _dispatch_root_queue_drain + 376
14 libdispatch.dylib              0x1b20b22c0 _dispatch_worker_thread2 + 128
15 libsystem_pthread.dylib        0x1b22e517c _pthread_wqthread + 472
16 libsystem_pthread.dylib        0x1b22e7cec start_wqthread + 4
```

hellais:
And `#78`

lorenzoPrimi:
CC: https://github.com/aryaxt/OCMapper/issues/69

PR #241, opened 2019-02-27
init shared instance once #223 #222

### LINKED ISSUE #223: Crash ObjectMapper.m line 163
Affects 10 users and is fabric `#62`.

```
Fatal Exception: NSInvalidArgumentException
0  CoreFoundation                 0x1a2408ec4 __exceptionPreprocess
1  libobjc.A.dylib                0x1a15d9a40 objc_exception_throw
2  CoreFoundation                 0x1a2321c24 -[NSOrderedSet initWithSet:copyItems:]
3  CoreFoundation                 0x1a240e7dc ___forwarding___
4  CoreFoundation                 0x1a241048c _CF_forwarding_prep_0
5  Foundation                     0x1a2e95cd8 NSClassFromString
6  ooniprobe                      0x1003d07f4 -[ObjectMapper processDictionaryFromObject:] (ObjectMapper.m:163)
7  ooniprobe                      0x1003d02bc -[ObjectMapper dictionaryFromObject:] (ObjectMapper.m:113)
8  ooniprobe                      0x1003d0a74 -[ObjectMapper processDictionaryFromObject:] (ObjectMapper.m:176)
9  ooniprobe                      0x1003d02bc -[ObjectMapper dictionaryFromObject:] (ObjectMapper.m:113)
10 ooniprobe                      0x1003cfa6c -[NSObject(ObjectMapper) dictionary] (NSObject+ObjectMapper.m:39)
11 ooniprobe                      0x100376704 -[Settings getSettingsDictionary] (Settings.m:23)
12 ooniprobe                      0x10038e424 __24-[MKNetworkTest runTest]_block_invoke.36 (MKNetworkTest.mm:40)
13 libdispatch.dylib              0x1a1e416c8 _dispatch_call_block_and_release
14 libdispatch.dylib              0x1a1e42484 _dispatch_client_callout
15 libdispatch.dylib              0x1a1de5490 _dispatch_queue_override_invoke
16 libdispatch.dylib              0x1a1df1a18 _dispatch_root_queue_drain
17 libdispatch.dylib              0x1a1df22c0 _dispatch_worker_thread2
18 libsystem_pthread.dylib        0x1a202517c _pthread_wqthread
19 libsystem_pthread.dylib        0x1a2027cec start_wqthread
```

lorenzoPrimi:
Fixes #223 #222

## Proposed Changes

  - Setting InCodeMappingProvider only once.
It is needed we use the shared instance as the function `dictionary` uses that sharedInstance
https://github.com/aryaxt/OCMapper/blob/master/OCMapper/Source/Categories/NSObject+ObjectMapper.m#L39

### DIFF
--- a/ooniprobe/AppDelegate.mm
+++ b/ooniprobe/AppDelegate.mm
@@ -25,6 +25,10 @@
     [[UINavigationBar appearance] setBarTintColor:[UIColor colorWithRGBHexString:color_blue5 alpha:1.0f]];
     [[UINavigationBar appearance] setTranslucent:FALSE];
 
+    InCodeMappingProvider *inCodeMappingProvider = [InCodeMappingProvider new];
+    [inCodeMappingProvider mapFromPropertyKey:@"ca_bundle_path" toDictionaryKey:@"net/ca_bundle_path" forClass:[Options class]];
+    [[ObjectMapper sharedInstance] setMappingProvider:inCodeMappingProvider];
+
     #ifdef RELEASE
     CrashlyticsKit.delegate = self;
     [Fabric with:@[[Crashlytics class]]];
--- a/ooniprobe/AppDelegate.mm
+++ b/ooniprobe/AppDelegate.mm
@@ -25,10 +25,8 @@
     [[UINavigationBar appearance] setBarTintColor:[UIColor colorWithRGBHexString:color_blue5 alpha:1.0f]];
     [[UINavigationBar appearance] setTranslucent:FALSE];
 
-    InCodeMappingProvider *inCodeMappingProvider = [InCodeMappingProvider new];
-    [inCodeMappingProvider mapFromPropertyKey:@"ca_bundle_path" toDictionaryKey:@"net/ca_bundle_path" forClass:[Options class]];
-    [[ObjectMapper sharedInstance] setMappingProvider:inCodeMappingProvider];
-
+    [self setMappingProvider];
+    
     #ifdef RELEASE
     CrashlyticsKit.delegate = self;
     [Fabric with:@[[Crashlytics class]]];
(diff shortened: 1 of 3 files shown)


lorenzoPrimi:
@bassosimone so is it correct to set the inCodeMappingProvider once and at the app start then use the shared instance when needed?

bassosimone:
@lorenzoPrimi yes that _feels_ more correct because it’s in a single threaded context.

In all honesty, we cannot proof that this fix _will_ make the crashes go away.

It has now occurred to me that different properties may have different attributes and perhaps there is one that allows overriding a property and not destroying the old pointed value (but I am in trouble to understand how that would work).

Since initialising a global object once feels more correct anyway, let’s try this diff and then see if the crashes reduce.

lorenzoPrimi:
I also initialize the settings object outside the background task to check if that would cause less memory issues. This PR is going in the 2.0.5 so we can check if it keeps crashing on #222 and #223

bassosimone:
I think it’s weird that you ask my review, I approve a set of changes, then you change that quite significantly and merge.

lorenzoPrimi:
the code it's only been duplicated into the foreground method as well as the app starting, no big changes at all

bassosimone:
On top of that, we discussed together a possibly simpler implementation of the same functionality which does not require the usage of any global object. I believe all the anxiety caused by checking whether a pointer is null or not is gone if we just avoid using global objects. The code will be smaller and simpler to understand. Why did you disregard that solution without even mentioning it? Is it wrong? If so, why?

lorenzoPrimi:
That is something that is not doable at the moment. I want to go for the easy and working solution, after 2.1 and having unit tests we can consider the other solution. 
In the meantime we will monitor fabric.

elioqoshi:
@lorenzoPrimi Your logic totally makes sense, that's not what @bassosimone was referring to however, rather the decision-making which you took on your own without confirming with @bassosimone (since you asked him in the first place). Technically that's okay to do but from a more human perspective we can do better.

lorenzoPrimi:
The method `- (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions`
is called once when the app starts, (or when it's opened fron killed state) it's used to setup basic things
while method `- (void)applicationWillEnterForeground:(UIApplication *)application`
is called whenever the app is opened from a backdroup state

the `inCodeMappingProvider` of the sharedInstance is setup in both methods (in case is not null) just to avoid losing that in case of memory low. It shouldn't happen but at the moment is still better than out current approach.

A method to modify the `ObjectMapper` class and not use a `sharedInstance` has been tried but I don't feel confident on using it now and as I'm working hard with Unit Tests I'd rather postpone that after the 2.1, after we have unit tests (and I have confidence with them) and based on what fabic tells us.

PR #257, opened 2019-03-30
sharkORM: use git repository rather than CocoaPods

lorenzoPrimi:
As the creator of SharkORM said, we were using an old version (2.3.62) who has a problem with migration and column removal : https://github.com/sharksync/sharkorm/issues/115

Unfortunately he has some problem with CocoaPods so he suggested to install the framework from the github repo for now as stated in this conversation with me https://github.com/sharksync/sharkorm/issues/27

In this branch I remove the CocoaPod and I install the latest framework v2.3.67 (https://github.com/sharksync/sharkorm/releases/tag/v2.3.67), furthermore I remove an unused column from the database, I tested and migration worked perfectly.


Podfile
```diff
@@ -9,7 +9,7 @@ target 'ooniprobe' do
     pod 'MBProgressHUD'
     pod 'DZNEmptyDataSet'
     pod 'lottie-ios'
-    pod 'SharkORM'
+#    pod 'SharkORM'
```
bassosimone: I believe we have the option to point the `pod` to the github repository rather than to a pod, which is what we do with `mkall-ios` currently. Can we try that as an option before vendording?

lorenzoPrimi:
I like this approach more. Did you also remove the .framework from the project?

lorenzoPrimi:
I approve this PR

bassosimone:
@lorenzoPrimi wrote:
>  furthermore I remove an unused column from the database

Oops, I think I dropped this part.

PR #265, opened 2019-04-17
Rename and refactor utility functions

lorenzoPrimi:


ooniprobe/Utility/TestUtility.mm
```diff
@@ -123,8 +123,24 @@ + (void)removeFile:(NSString*)fileName {
     }
 }
 
-+ (long)numberOfTest:(NSString*)testName{
```
lorenzoPrimi: function not used

PR #268, opened 2019-04-17
DB functions to check not upload results

lorenzoPrimi:


OONIProbeUnitTests/ResultTest.m
```diff
@@ -0,0 +1,58 @@
...
+
+- (void)testUploaded {
+    //Measurement uploaded but report_id null, isEveryMeasurementUploaded false
+    measurement.is_uploaded = true;
+    [measurement save];
+    XCTAssert(![result isEveryMeasurementUploaded]);
+}
+
+- (void)testUploadedAndReport {
+    //Measurement uploaded and wirh report_id, isEveryMeasurementUploaded true
```
bassosimone: wirh => with

OONIProbeUnitTests/ResultTest.m
```diff
@@ -0,0 +1,58 @@
...
+    measurement.result_id = result;
+    [measurement save];
+}
+
+- (void)tearDown {
+    // Put teardown code here. This method is called after the invocation of each test method in the class.
+}
+
+- (void)testNoMeasurements {
+    //No measurement, isEveryMeasurementUploaded true
```
bassosimone: This is now wrong: there is a measurement in the database

lorenzoPrimi: read it again, i use [Result new] not the previous initialized result

bassosimone: So, to be really really really pick, I'd drop the global object and just write something like:

```ObjectiveC
- (void)testNoMeasurements {
  Result *result = [Result new];
  Measurement *measurement = [Measurement new];
  measurement.result_id = result;
  [measurement save];
  // Testing code...
}
```

Also, I still maintain that `testNoMeasurement` is not the correct name. You clearly have one saved measurement in the database, rather than no measurement. Please:

1. drop the globals

2. make it obvious what is the state of each test case (implicit is in general bad and here it seems very bad to the point that, unless I'm seriously mistaken, we have a "no measurement" test that contains measurements).

3. as an additional mechanism for robustness, please count that the number of all measurements is consistent with the number of measurements you actually inserted (so I don't have to guess but the code testifies that).

Thanks <3

bassosimone: > You clearly have one saved measurement in the database, rather than no measurement.

This is fake news. The code is actually different than I though. But I guess this is also because I didn't expect the code to do what it is doing. Anyway, my apologies.

PR #275, opened 2019-05-20
Retry upload alert

lorenzoPrimi:
In `UploadFooterViewController.h` not the `uploadMeasurements` method gets an array of measurements and the index where to start from.
If an upload fails and the user press Retry, the index will be the current failed upload measurement.

The change in the `Result.h` file in the NOT_UPLOADED_QUERY is due to the face we don't want to 
show the not uploaded banner if there are failed or re_run tests not uploaded.


ooniprobe/View/TestResults/Footer/UploadFooterViewController.m
```diff
@@ -65,45 +65,46 @@ -(void)uploadResult{
...
-        [self uploadSingleMeasurement:self.measurement];
-        dispatch_async(dispatch_get_main_queue(), ^{
-            [MBProgressHUD HUDForView:self.navigationController.view].progress = 1;
-        });
+        [self uploadMeasurements:[NSArray arrayWithObject:self.measurement]];
     }
 }
 
--(void)uploadMeasurements:(SRKResultSet *)notUploaded{
+-(void)uploadMeasurements:(NSArray *)notUploaded {
```
lorenzoPrimi: `SRKResultSet` is a subclass of `NSArray`

ooniprobe/View/TestResults/Footer/UploadFooterViewController.m
```diff
@@ -65,45 +65,46 @@ -(void)uploadResult{
...
+    [self uploadMeasurements:notUploaded startAt:0];
+}
+
+-(void)uploadMeasurements:(NSArray *)notUploaded startAt:(int)idx{
     if ([notUploaded count] == 0) return;
```
bassosimone: Please, use `<= 0` here to also cover the unlikely but possible case where the index is negative.

ooniprobe/View/TestResults/Footer/UploadFooterViewController.m
```diff
@@ -65,45 +65,46 @@ -(void)uploadResult{
...
-            [MBProgressHUD HUDForView:self.navigationController.view].label.text = NSLocalizedFormatString(@"Modal.ResultsNotUploaded.Uploading", [NSString stringWithFormat:@"%d/%ld", done, [notUploaded count]]);
-        });
-        [self uploadSingleMeasurement:currentMeasurement];
-        progress += measurementValue;
-        done++;
+    while (idx < [notUploaded count]){
+        Measurement *currentMeasurement = [notUploaded objectAtIndex:idx];
         dispatch_async(dispatch_get_main_queue(), ^{
-            [MBProgressHUD HUDForView:self.navigationController.view].progress = progress;
+            [MBProgressHUD HUDForView:self.navigationController.view].label.text = NSLocalizedFormatString(@"Modal.ResultsNotUploaded.Uploading", [NSString stringWithFormat:@"%d/%ld", idx+1, [notUploaded count]]);
```
bassosimone: Can you break the line after `NSLocalizedFormatString` to increase the diff readability?

ooniprobe/Model/Database/Result.h
```diff
@@ -2,7 +2,7 @@
 #import <SharkORM/SharkORM.h>
 #import "JsonResult.h"
 #import "Measurement.h"
-#define NOT_UPLOADED_QUERY @"is_failed = 0 AND (is_uploaded = 0 || report_id IS NULL)"
+#define NOT_UPLOADED_QUERY @"is_failed = 0 AND is_rerun = 0 AND is_done = 1 AND (is_uploaded = 0 || report_id IS NULL)"
```
bassosimone: Did you add these checks because the underlying set could mutate as we process it?

ooniprobe/View/TestResults/Footer/UploadFooterViewController.m
bassosimone: Since this may also surprise others in the future, maybe add a comment in the code?

ooniprobe/View/TestResults/Footer/UploadFooterViewController.m
```diff
@@ -118,6 +119,20 @@ -(void)uploadSingleMeasurement:(Measurement*)measurement{
         [measurement setReport_id:[results updatedReportID]];
         [measurement save];
     }
+    return [results good];
 }
 
+-(void)showRetryPopup:(NSArray *)notUploaded :(int)start{
```
bassosimone: Here it would be better to remove the space between `notUploaded` and `:`

ooniprobe/View/TestResults/Footer/UploadFooterViewController.m
```diff
@@ -118,6 +119,20 @@ -(void)uploadSingleMeasurement:(Measurement*)measurement{
...
+    UIAlertAction* okButton = [UIAlertAction
+                               actionWithTitle:NSLocalizedString(@"Modal.Retry", nil)
+                               style:UIAlertActionStyleDefault
+                               handler:^(UIAlertAction * action) {
```
bassosimone: I'd remove the space between `*` and `action`

ooniprobe/View/TestResults/Footer/UploadFooterViewController.m
lorenzoPrimi: An NSArray count can't be negative by implementation. It return an unsigned NSInteger

ooniprobe/Model/Database/Result.h
lorenzoPrimi: No, just because we don't care about re_run tests and of course the unfinished ones (they will never show up in results)

ooniprobe/View/TestResults/Footer/UploadFooterViewController.m
lorenzoPrimi: It's obj-c syntax. I can't remove the space, I can only add a name for the param, will do that.

ooniprobe/View/TestResults/Footer/UploadFooterViewController.m
bassosimone: Gotcha

ooniprobe/Model/Database/Result.h
bassosimone: Gotcha

ooniprobe/View/TestResults/Footer/UploadFooterViewController.m
bassosimone: ah, wow, I completely missed this aspect. Yes, please!

ooniprobe/View/TestResults/Footer/UploadFooterViewController.m
```diff
@@ -65,45 +65,48 @@ -(void)uploadResult{
...
-        });
-        [self uploadSingleMeasurement:currentMeasurement];
-        progress += measurementValue;
-        done++;
+    while (idx < [notUploaded count]){
+        Measurement *currentMeasurement = [notUploaded objectAtIndex:idx];
         dispatch_async(dispatch_get_main_queue(), ^{
-            [MBProgressHUD HUDForView:self.navigationController.view].progress = progress;
+            [MBProgressHUD HUDForView:self.navigationController.view].label.text =
+            NSLocalizedFormatString(@"Modal.ResultsNotUploaded.Uploading", [NSString stringWithFormat:@"%d/%ld", idx+1, [notUploaded count]]);
```
bassosimone: I meant to use this formatting, which is less suprising. Please apply it:

```ObjectiveC
[MBProgressHUD HUDForView:self.navigationController.view].label.text = NSLocalizedFormatString(
    @"Modal.ResultsNotUploaded.Uploading",
    [NSString stringWithFormat:@"%d/%ld", idx+1, [notUploaded count]]);
```

lorenzoPrimi: I don't like to cut after NSLocalizedFormatString(
any other solution?

bassosimone:
```ObjectiveC
[MBProgressHUD HUDForView:self.navigationController.view].label.text =
    NSLocalizedFormatString(@"Modal.ResultsNotUploaded.Uploading",
        [NSString stringWithFormat:@"%d/%ld", idx+1, [notUploaded count]]);
```

?

ooniprobe/View/TestResults/Footer/UploadFooterViewController.m
```diff
@@ -90,7 +90,8 @@ -(void)uploadMeasurements:(NSArray *)notUploaded startAt:(int)idx{
         Measurement *currentMeasurement = [notUploaded objectAtIndex:idx];
         dispatch_async(dispatch_get_main_queue(), ^{
             [MBProgressHUD HUDForView:self.navigationController.view].label.text =
-            NSLocalizedFormatString(@"Modal.ResultsNotUploaded.Uploading", [NSString stringWithFormat:@"%d/%ld", idx+1, [notUploaded count]]);
+            NSLocalizedFormatString(@"Modal.ResultsNotUploaded.Uploading",
```
bassosimone: Please also indent `NSLocalizedFormatString` four spaces on the right, otherwise it's not apparent that this line is a continuation from the previous line.

OONIProbeUnitTests/ResultTest.m
```diff
@@ -32,17 +32,18 @@ - (void)testNoMeasurements {
 }
 
 - (void)testNotUploaded {
-    //One measurement, isEveryMeasurementUploaded false
+    //One measurement, with is_done false, isEveryMeasurementUploaded true
     Result *result = [Result new];
-    Measurement *measurement = [self addMeasurement:result];
-    XCTAssert(![result isEveryMeasurementUploaded]);
+    [self addMeasurement:result];
+    XCTAssert([result isEveryMeasurementUploaded]);
```
bassosimone: I do not get exactly why the code change broke the original test. Can you explain?

OONIProbeUnitTests/ResultTest.m
```diff
@@ -32,17 +32,18 @@ - (void)testNoMeasurements {
...
 
 - (void)testUploaded {
     //Measurement uploaded but report_id null, isEveryMeasurementUploaded false
     Result *result = [Result new];
     Measurement *measurement = [self addMeasurement:result];
     measurement.is_uploaded = true;
+    measurement.is_done = true;
     [measurement save];
     XCTAssert(![result isEveryMeasurementUploaded]);
 }
```
bassosimone: Can you also explain why here the test needed to be changed?

ooniprobe/Model/Database/Result.h
bassosimone: One more question: why don't we care about rerun tests?

ooniprobe/View/TestResults/Footer/UploadFooterViewController.m
```diff
@@ -65,45 +65,49 @@ -(void)uploadResult{
...
     float measurementValue = 1.0/[notUploaded count];
-    int done = 1;
-    for (Measurement *currentMeasurement in notUploaded){
-        dispatch_async(dispatch_get_main_queue(), ^{
-            [MBProgressHUD HUDForView:self.navigationController.view].label.text = NSLocalizedFormatString(@"Modal.ResultsNotUploaded.Uploading", [NSString stringWithFormat:@"%d/%ld", done, [notUploaded count]]);
-        });
-        [self uploadSingleMeasurement:currentMeasurement];
-        progress += measurementValue;
-        done++;
+    while (idx < [notUploaded count]){
```
bassosimone: I think the type of `idx` should be the same type returnd by `[notUploaded count]`

ooniprobe/View/TestResults/Footer/UploadFooterViewController.m
```diff
@@ -65,45 +65,49 @@ -(void)uploadResult{
...
+            progress += measurementValue;
+            idx++;
+            dispatch_async(dispatch_get_main_queue(), ^{
+                [MBProgressHUD HUDForView:self.navigationController.view].progress = progress;
+            });
+        }
+        else {
+            [self showRetryPopup:notUploaded startAt:idx];
+            return;
+        }
```
bassosimone: I believe the complexity of the function would be reduced if you'd write:

```ObjectiveC
  if (![self uploadMeasurement:currentMeasurement]) {
    [self showRetryPopup:notUploaded startAt:idx];
    return;
  }
 // other code
```

ooniprobe/Model/Database/Result.h
lorenzoPrimi: because a test who is re ran is usually a failed test and it has no json.
Furthermore is not visible in the list so it wil never get the upload bottom banner

bassosimone: Right. Is it correct to say, though, that in principle a re-run test may not be failed? What happens in that case? Is the output file overwritten? Is this a corner case we need to care (maybe just by opening an issue and dealing with it in the future)?

ooniprobe/View/TestResults/Footer/UploadFooterViewController.m
lorenzoPrimi: it is the same, they are both int

OONIProbeUnitTests/ResultTest.m
lorenzoPrimi: because now we don't care about tests who are not done. 
This is how the test should have been written since the beginning

OONIProbeUnitTests/ResultTest.m
lorenzoPrimi: a test who is not done can't be reuploaded

ooniprobe/Model/Database/Result.h
lorenzoPrimi: a test who has been re ran is failed.
Everything will be overwritten in the moment it will be re run, except the db row (a new one is created)

ooniprobe/View/TestResults/Footer/UploadFooterViewController.m
bassosimone: Are you sure? NSUInteger strikes me as an _unsigned_ integer:

https://developer.apple.com/documentation/foundation/nsarray/1409982-count?language=objc

ooniprobe/Model/Database/Result.h
bassosimone: Right, we can only re-run a test that failed. We may offer used to re-run a test in general in the future. When we will do that, we will probably need to think at the implications.

ooniprobe/View/TestResults/Footer/UploadFooterViewController.m
lorenzoPrimi: I don't really know good these obj-c types but surely I will rename it

OONIProbeUnitTests/ResultTest.m
bassosimone: So, the function should probably be renamed `testNotDone`?

OONIProbeUnitTests/ResultTest.m
bassosimone: ack

PR #276, opened 2019-05-22
Set timeout for upload tests

lorenzoPrimi:

### DIFF
--- a/ooniprobe/Utility/TestUtility.mm
+++ b/ooniprobe/Utility/TestUtility.mm
@@ -143,4 +143,11 @@
         [data writeToFile:fileName atomically:YES];
     }
 }
+
++ (NSUInteger)makeTimeout:(NSUInteger)bytes{
+    //Timeout dependent on the body size considering a minimum upload speed of 12.5 kbit/s + 10 s
+    NSUInteger timeout = bytes / 2000 + 10;
+    return timeout;
+}
+
 @end
--- a/ooniprobe/Utility/TestUtility.mm
+++ b/ooniprobe/Utility/TestUtility.mm
@@ -145,7 +145,7 @@
 }
 
 + (NSUInteger)makeTimeout:(NSUInteger)bytes{
-    //Timeout dependent on the body size considering a minimum upload speed of 12.5 kbit/s + 10 s
+    //Timeout dependent on the body size considering a minimum upload speed of 16 kbit/s + 10 s
     NSUInteger timeout = bytes / 2000 + 10;
     return timeout;
 }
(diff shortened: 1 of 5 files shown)


ooniprobe/View/TestResults/Footer/UploadFooterViewController.m
```diff
@@ -6,6 +6,7 @@
 #import "TestUtility.h"
 #import "MBProgressHUD.h"
 #import "VersionUtility.h"
+#define STANDARD_TIMEOUT 30
```
bassosimone: I would call this minimum timeout

ooniprobe/View/TestResults/Footer/UploadFooterViewController.m
```diff
@@ -107,7 +108,11 @@ -(void)uploadMeasurements:(NSArray *)notUploaded startAt:(NSUInteger)idx{
 
 -(BOOL)uploadMeasurement:(Measurement*)measurement{
     NSString *content = [TestUtility getUTF8FileContent:[measurement getReportFile]];
+    NSUInteger bytes = [content lengthOfBytesUsingEncoding:NSUTF8StringEncoding];
     MKCollectorResubmitSettings *settings = [[MKCollectorResubmitSettings alloc] init];
+    //Timeout dependent on the body size considering a minimum upload speed of 12.5 kbit/s + 10 s
+    NSUInteger timeout = bytes * 8 / (12.5 * 1000) + 10;
+    [settings setTimeout:timeout < STANDARD_TIMEOUT ? STANDARD_TIMEOUT : timeout];
```
bassosimone: Strictly speaking we should check that bytes * 8 does not overflow. It would be better to avoid multiplying by `8` and represent the speed in `kbit/s` so we cannot overflow.

bassosimone: I would like this functionality to be a static method and I would like to see tests for some corner cases plus a common case.

ooniprobe/Utility/TestUtility.mm
```diff
@@ -143,4 +143,11 @@ + (void)writeString:(NSString*)str toFile:(NSString*)fileName{
         [data writeToFile:fileName atomically:YES];
     }
 }
+
++ (NSUInteger)makeTimeout:(NSUInteger)bytes{
+    //Timeout dependent on the body size considering a minimum upload speed of 12.5 kbit/s + 10 s
```
bassosimone: I think this comment is now inaccurate because we rounded. I think that's 16 Kbit/s.

PR #293, opened 2019-06-23
Add API to fetch measurement URL

lorenzoPrimi:
This should do the trick, I just don't understand why we need the URL. 
Can't we call this function for non-webconnectivity tests?

### DIFF
--- a/OONIProbeUnitTests/APITest.m
+++ b/OONIProbeUnitTests/APITest.m
@@ -1,0 +1,40 @@
+#import <XCTest/XCTest.h>
+#import "Measurement.h"
+#define EXISTING_REPORT_ID @"20190113T202156Z_AS327931_CgoC3KbgM6zKajvIIt1AxxybJ1HbjwwWJjsJnlxy9rpcGY54VH"
+#define NONEXISTING_REPORT_ID @"EMPTY"
+
+@interface APITest : XCTestCase
+
+@end
+
+@implementation APITest
+
+- (void)setUp {
+    // Put setup code here. This method is called before the invocation of each test method in the class.
+}
+
+- (void)tearDown {
+    // Put teardown code here. This method is called after the invocation of each test method in the class.
+}
+
+- (void)testExisting {
+    Measurement *measurement = [Measurement new];
+    measurement.report_id = EXISTING_REPORT_ID;
+    [measurement getExplorerUrl:^(NSString *measurement_url){
+        XCTAssert(true);
+    } onError:^(NSError *error){
+        XCTAssert(false);
+    }];
+}
+
+- (void)testNonExisting {
+    Measurement *measurement = [Measurement new];
+    measurement.report_id = NONEXISTING_REPORT_ID;
+    [measurement getExplorerUrl:^(NSString *measurement_url){
+        XCTAssert(false);
+    } onError:^(NSError *error){
+        XCTAssert(true);
+    }];
+}
+
+@end
--- a/OONIProbeUnitTests/APITest.m
+++ b/OONIProbeUnitTests/APITest.m
@@ -1,5 +1,6 @@
 #import <XCTest/XCTest.h>
 #import "Measurement.h"
+#import "TestUtility.h"
 #define EXISTING_REPORT_ID @"20190113T202156Z_AS327931_CgoC3KbgM6zKajvIIt1AxxybJ1HbjwwWJjsJnlxy9rpcGY54VH"
 #define NONEXISTING_REPORT_ID @"EMPTY"
 
@@ -34,6 +35,14 @@
         XCTAssert(false);
     } onError:^(NSError *error){
         XCTAssert(true);
+    }];
+}
+
+-(void)testDownloadUrls{
+    [TestUtility downloadUrls:^(NSArray *urls) {
+        XCTAssert(true);
+    } onError:^(NSError *error) {
+        XCTAssert(false);
     }];
 }
 
--- a/OONIProbeUnitTests/APITest.m
+++ b/OONIProbeUnitTests/APITest.m
@@ -19,6 +19,7 @@
 }
 
 - (void)testExisting {
+    //TODO test getExplorerUrlCallback with custom NSData
     Measurement *measurement = [Measurement new];
     measurement.report_id = EXISTING_REPORT_ID;
     [measurement getExplorerUrl:^(NSString *measurement_url){
@@ -39,6 +40,7 @@
 }
 
 -(void)testDownloadUrls{
+    //TODO test downloadUrlsCallback with custom NSData
     [TestUtility downloadUrls:^(NSArray *urls) {
         XCTAssert(true);
     } onError:^(NSError *error) {
--- a/ooniprobe/Model/Database/Measurement.h
+++ b/ooniprobe/Model/Database/Measurement.h
@@ -38,5 +38,5 @@
 -(void)save;
 -(void)setReRun;
 -(void)deleteObject;
--(void)getExplorerUrl:(void (^)(NSString*))successcb error:(void (^)(NSError*))errorcb;
+-(void)getExplorerUrl:(void (^)(NSString*))successcb onError:(void (^)(NSError*))errorcb;
 @end
--- a/ooniprobe/Model/Database/Measurement.h
+++ b/ooniprobe/Model/Database/Measurement.h
@@ -38,5 +38,5 @@
 -(void)save;
 -(void)setReRun;
 -(void)deleteObject;
--(void)getExplorerUrl:(void (^)(NSString *))measurement_url;
+-(void)getExplorerUrlSuccessCallback:(void (^)(NSString*))success errorCallback:(void (^)(NSError*))error;
 @end
--- a/ooniprobe/Model/Database/Measurement.h
+++ b/ooniprobe/Model/Database/Measurement.h
@@ -36,5 +36,5 @@
 -(void)save;
 -(void)setReRun;
 -(void)deleteObject;
-
+-(void)getExplorerUrl:(void (^)(NSString *))measurement_url;
 @end
--- a/ooniprobe/Model/Database/Measurement.h
+++ b/ooniprobe/Model/Database/Measurement.h
@@ -38,5 +38,5 @@
 -(void)save;
 -(void)setReRun;
 -(void)deleteObject;
--(void)getExplorerUrlSuccessCallback:(void (^)(NSString*))success errorCallback:(void (^)(NSError*))error;
+-(void)getExplorerUrl:(void (^)(NSString*))successcb error:(void (^)(NSError*))errorcb;
 @end
--- a/ooniprobe/Model/Database/Measurement.mm
+++ b/ooniprobe/Model/Database/Measurement.mm
@@ -117,8 +117,7 @@
              return;
          }
          NSArray *resultsArray = [dic objectForKey:@"results"];
-         //TODO symbolize somehow empty array
-         if ([resultsArray count] <= 0) {
+         if ([resultsArray count] <= 0 || ![[resultsArray objectAtIndex:0] objectForKey:@"measurement_url"]) {
              errorcb([NSError errorWithDomain:@"io.ooni.api"
                                          code:ERR_JSON_EMPTY
                                      userInfo:@{NSLocalizedDescriptionKey:@"Error.JsonEmpty"
--- a/ooniprobe/Model/Database/Measurement.mm
+++ b/ooniprobe/Model/Database/Measurement.mm
@@ -96,7 +96,7 @@
     [self remove];
 }
 
--(void)getExplorerUrl:(void (^)(NSString*))successcb error:(void (^)(NSError*))errorcb {
+-(void)getExplorerUrl:(void (^)(NSString*))successcb onError:(void (^)(NSError*))errorcb {
     NSLog(@"%@ getExplorerUrl",self.Id);
     NSMutableString *path = [NSMutableString stringWithFormat:@"https://api.ooni.io/api/v1/measurements?report_id=%@",
                       self.report_id];
@@ -120,7 +120,7 @@
          //TODO symbolize somehow empty array
          if ([resultsArray count] <= 0) {
              errorcb([NSError errorWithDomain:@"io.ooni.api"
-                                         code:100
+                                         code:JSON_EMPTY_CODE
                                      userInfo:@{NSLocalizedDescriptionKey:@"Error.JsonEmpty"
                                                 }]);
              return;
--- a/ooniprobe/Model/Database/Measurement.mm
+++ b/ooniprobe/Model/Database/Measurement.mm
@@ -97,13 +97,17 @@
 }
 
 -(void)getExplorerUrl:(void (^)(NSString*))successcb onError:(void (^)(NSError*))errorcb {
-    NSLog(@"%@ getExplorerUrl",self.Id);
-    NSMutableString *path = [NSMutableString stringWithFormat:@"https://api.ooni.io/api/v1/measurements?report_id=%@",
-                      self.report_id];
+    NSURLComponents *components = [[NSURLComponents alloc] init];
+    components.scheme = @"https";
+    components.host = @"api.ooni.io";
+    components.path = @"/api/v1/measurements";
     if ([self.test_name isEqualToString:@"web_connectivity"])
-        [path appendString:[NSString stringWithFormat:@"&input=%@",
-                            self.url_id.url]];
-    NSURL *url = [NSURL URLWithString:path];
+        components.query = [NSString stringWithFormat:@"report_id=%@&input=%@",
+                            self.report_id, self.url_id.url];
+    else
+        components.query = [NSString stringWithFormat:@"report_id=%@",
+                        self.report_id];
+    NSURL *url = components.URL;
     NSURLSessionDataTask *downloadTask = [[NSURLSession sharedSession]
      dataTaskWithURL:url
      completionHandler:^(NSData *data, NSURLResponse *response, NSError *error) {
--- a/ooniprobe/Model/Database/Measurement.mm
+++ b/ooniprobe/Model/Database/Measurement.mm
@@ -104,7 +104,7 @@
     NSURLQueryItem *reportIdItem = [NSURLQueryItem
                                     queryItemWithName:@"report_id"
                                     value:self.report_id];
-
+    //web_connectivity is the only test using input for now
     if ([self.test_name isEqualToString:@"web_connectivity"]){
         NSURLQueryItem *urlItem = [NSURLQueryItem
                                    queryItemWithName:@"input"
--- a/ooniprobe/Model/Database/Measurement.mm
+++ b/ooniprobe/Model/Database/Measurement.mm
@@ -96,7 +96,7 @@
     [self remove];
 }
 
--(void)getExplorerUrlSuccessCallback:(void (^)(NSString*))success errorCallback:(void (^)(NSError*))error {
+-(void)getExplorerUrl:(void (^)(NSString*))successcb error:(void (^)(NSError*))errorcb {
     NSLog(@"%@ getExplorerUrl",self.Id);
     NSMutableString *path = [NSMutableString stringWithFormat:@"https://api.ooni.io/api/v1/measurements?report_id=%@",
                       self.report_id];
@@ -107,21 +107,21 @@
     NSURLSessionDataTask *downloadTask =
     [[NSURLSession sharedSession]
      dataTaskWithURL:url
-     completionHandler:^(NSData *data, NSURLResponse *response, NSError *nserror) {
+     completionHandler:^(NSData *data, NSURLResponse *response, NSError *error) {
          if (nserror != nil)
-             error(nserror);
+             errorcb(error);
          else {
-             NSDictionary *dic = [NSJSONSerialization JSONObjectWithData:data options:0 error:&nserror];
-             if (nserror != nil)
-                 error(nserror);
+             NSDictionary *dic = [NSJSONSerialization JSONObjectWithData:data options:0 error:&error];
+             if (error != nil)
+                 errorcb(error);
              NSArray *resultsArray = [dic objectForKey:@"results"];
              //TODO symbolize somehow empty array
              if ([resultsArray count] == 0)
-                 error([NSError errorWithDomain:@"io.ooni.api"
+                 errorcb([NSError errorWithDomain:@"io.ooni.api"
                                            code:100
                                        userInfo:@{NSLocalizedDescriptionKey:@"Error.JsonEmpty"
                                                   }];);
-             success([[resultsArray objectAtIndex:0] objectForKey:@"measurement_url"]);
+             successcb([[resultsArray objectAtIndex:0] objectForKey:@"measurement_url"]);
          }
     }];
     [downloadTask resume];
--- a/ooniprobe/Model/Database/Measurement.mm
+++ b/ooniprobe/Model/Database/Measurement.mm
@@ -108,19 +108,25 @@
     [[NSURLSession sharedSession]
      dataTaskWithURL:url
      completionHandler:^(NSData *data, NSURLResponse *response, NSError *error) {
-         if (error != nil)
+         if (error != nil){
              errorcb(error);
+             return;
+         }
          else {
              NSDictionary *dic = [NSJSONSerialization JSONObjectWithData:data options:0 error:&error];
-             if (error != nil)
+             if (error != nil){
                  errorcb(error);
+                 return;
+             }
              NSArray *resultsArray = [dic objectForKey:@"results"];
              //TODO symbolize somehow empty array
-             if ([resultsArray count] > 0)
+             if ([resultsArray count] > 0){
                  errorcb([NSError errorWithDomain:@"io.ooni.api"
                                            code:100
                                        userInfo:@{NSLocalizedDescriptionKey:@"Error.JsonEmpty"
                                                   }]);
+                 return;
+             }
              successcb([[resultsArray objectAtIndex:0] objectForKey:@"measurement_url"]);
          }
     }];
--- a/ooniprobe/Model/Database/Measurement.mm
+++ b/ooniprobe/Model/Database/Measurement.mm
@@ -120,7 +120,7 @@
          //TODO symbolize somehow empty array
          if ([resultsArray count] <= 0) {
              errorcb([NSError errorWithDomain:@"io.ooni.api"
-                                         code:JSON_EMPTY_CODE
+                                         code:ERR_JSON_EMPTY
                                      userInfo:@{NSLocalizedDescriptionKey:@"Error.JsonEmpty"
                                                 }]);
              return;
--- a/ooniprobe/Model/Database/Measurement.mm
+++ b/ooniprobe/Model/Database/Measurement.mm
@@ -101,12 +101,15 @@
     components.scheme = @"https";
     components.host = @"api.ooni.io";
     components.path = @"/api/v1/measurements";
-    if ([self.test_name isEqualToString:@"web_connectivity"])
-        components.query = [NSString stringWithFormat:@"report_id=%@&input=%@",
-                            self.report_id, self.url_id.url];
+    NSURLQueryItem *reportIdItem = [NSURLQueryItem queryItemWithName:@"report_id" value:self.report_id];
+
+    if ([self.test_name isEqualToString:@"web_connectivity"]){
+        NSURLQueryItem *urlItem = [NSURLQueryItem queryItemWithName:@"input" value:self.url_id.url];
+        components.queryItems = @[ reportIdItem, urlItem ];
+    }
     else
-        components.query = [NSString stringWithFormat:@"report_id=%@",
-                        self.report_id];
+        components.queryItems = @[ reportIdItem ];
+
     NSURL *url = components.URL;
     NSURLSessionDataTask *downloadTask = [[NSURLSession sharedSession]
      dataTaskWithURL:url
--- a/ooniprobe/Model/Database/Measurement.mm
+++ b/ooniprobe/Model/Database/Measurement.mm
@@ -101,10 +101,14 @@
     components.scheme = @"https";
     components.host = @"api.ooni.io";
     components.path = @"/api/v1/measurements";
-    NSURLQueryItem *reportIdItem = [NSURLQueryItem queryItemWithName:@"report_id" value:self.report_id];
+    NSURLQueryItem *reportIdItem = [NSURLQueryItem
+                                    queryItemWithName:@"report_id"
+                                    value:self.report_id];
 
     if ([self.test_name isEqualToString:@"web_connectivity"]){
-        NSURLQueryItem *urlItem = [NSURLQueryItem queryItemWithName:@"input" value:self.url_id.url];
+        NSURLQueryItem *urlItem = [NSURLQueryItem
+                                   queryItemWithName:@"input"
+                                   value:self.url_id.url];
         components.queryItems = @[ reportIdItem, urlItem ];
     }
     else
--- a/ooniprobe/Model/Database/Measurement.mm
+++ b/ooniprobe/Model/Database/Measurement.mm
@@ -108,28 +108,26 @@
     [[NSURLSession sharedSession]
      dataTaskWithURL:url
      completionHandler:^(NSData *data, NSURLResponse *response, NSError *error) {
-         if (error != nil){
+         if (error != nil) {
              errorcb(error);
              return;
          }
-         else {
-             NSDictionary *dic = [NSJSONSerialization JSONObjectWithData:data options:0 error:&error];
-             if (error != nil){
-                 errorcb(error);
-                 return;
-             }
-             NSArray *resultsArray = [dic objectForKey:@"results"];
-             //TODO symbolize somehow empty array
-             if ([resultsArray count] > 0){
-                 errorcb([NSError errorWithDomain:@"io.ooni.api"
-                                           code:100
-                                       userInfo:@{NSLocalizedDescriptionKey:@"Error.JsonEmpty"
-                                                  }]);
-                 return;
-             }
-             successcb([[resultsArray objectAtIndex:0] objectForKey:@"measurement_url"]);
+         NSDictionary *dic = [NSJSONSerialization JSONObjectWithData:data options:0 error:&error];
+         if (error != nil) {
+             errorcb(error);
+             return;
          }
-    }];
+         NSArray *resultsArray = [dic objectForKey:@"results"];
+         //TODO symbolize somehow empty array
+         if ([resultsArray count] <= 0) {
+             errorcb([NSError errorWithDomain:@"io.ooni.api"
+                                         code:100
+                                     userInfo:@{NSLocalizedDescriptionKey:@"Error.JsonEmpty"
+                                                }]);
+             return;
+         }
+         successcb([[resultsArray objectAtIndex:0] objectForKey:@"measurement_url"]);
+     }];
     [downloadTask resume];
 }
 
--- a/ooniprobe/Model/Database/Measurement.mm
+++ b/ooniprobe/Model/Database/Measurement.mm
@@ -118,26 +118,35 @@
     NSURLSessionDataTask *downloadTask = [[NSURLSession sharedSession]
      dataTaskWithURL:url
      completionHandler:^(NSData *data, NSURLResponse *response, NSError *error) {
-         if (error != nil) {
-             errorcb(error);
-             return;
-         }
-         NSDictionary *dic = [NSJSONSerialization JSONObjectWithData:data options:0 error:&error];
-         if (error != nil) {
-             errorcb(error);
-             return;
-         }
-         NSArray *resultsArray = [dic objectForKey:@"results"];
-         if ([resultsArray count] <= 0 || ![[resultsArray objectAtIndex:0] objectForKey:@"measurement_url"]) {
-             errorcb([NSError errorWithDomain:@"io.ooni.api"
-                                         code:ERR_JSON_EMPTY
-                                     userInfo:@{NSLocalizedDescriptionKey:@"Error.JsonEmpty"
-                                                }]);
-             return;
-         }
-         successcb([[resultsArray objectAtIndex:0] objectForKey:@"measurement_url"]);
+         [self getExplorerUrlCallback:data response:response error:error
+                          onSuccess:successcb onError:errorcb];
      }];
     [downloadTask resume];
 }
 
+- (void)getExplorerUrlCallback:(NSData *)data
+                    response:(NSURLResponse *)response
+                       error:(NSError *)error
+                   onSuccess:(void (^)(NSString*))successcb
+                     onError:(void (^)(NSError*))errorcb {
+    if (error != nil) {
+        errorcb(error);
+        return;
+    }
+    NSDictionary *dic = [NSJSONSerialization JSONObjectWithData:data options:0 error:&error];
+    if (error != nil) {
+        errorcb(error);
+        return;
+    }
+    NSArray *resultsArray = [dic objectForKey:@"results"];
+    if ([resultsArray count] <= 0 || ![[resultsArray objectAtIndex:0] objectForKey:@"measurement_url"]) {
+        errorcb([NSError errorWithDomain:@"io.ooni.api"
+                                    code:ERR_JSON_EMPTY
+                                userInfo:@{NSLocalizedDescriptionKey:@"Error.JsonEmpty"
+                                           }]);
+        return;
+    }
+    successcb([[resultsArray objectAtIndex:0] objectForKey:@"measurement_url"]);
+
+}
 @end
--- a/ooniprobe/Model/Database/Measurement.mm
+++ b/ooniprobe/Model/Database/Measurement.mm
@@ -117,7 +117,10 @@
              NSArray *resultsArray = [dic objectForKey:@"results"];
              //TODO symbolize somehow empty array
              if ([resultsArray count] == 0)
-                 error(nil);
+                 error([NSError errorWithDomain:@"io.ooni.api"
+                                           code:100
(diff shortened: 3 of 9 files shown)


ooniprobe/Model/Database/Measurement.mm
```diff
@@ -96,4 +96,24 @@ -(void)deleteObject{
     [self remove];
 }
 
+- (void)getExplorerUrl:(void (^)(NSString *))measurement_url {
+    NSString *path = [NSString stringWithFormat:@"https://api.ooni.io/api/v1/measurements?report_id=%@&input=%@", self.report_id, self.url_id.url];
```
bassosimone: Can you wrap this long line?

ooniprobe/Model/Database/Measurement.mm
```diff
@@ -96,4 +96,24 @@ -(void)deleteObject{
...
+    NSURL *url = [NSURL URLWithString:path];
+    NSURLSessionDataTask *downloadTask = [[NSURLSession sharedSession] dataTaskWithURL:url completionHandler:^(NSData *data, NSURLResponse *response, NSError *error) {
```
bassosimone: Can you wrap this long line?

ooniprobe/Model/Database/Measurement.mm
```diff
@@ -96,4 +96,24 @@ -(void)deleteObject{
...
+            NSArray *resultsArray = [dic objectForKey:@"results"];
+            if ([resultsArray count] == 0)
+                measurement_url(nil);
+            measurement_url([[resultsArray objectAtIndex:0] objectForKey:@"measurement_url"]);
+        }
+        else {
+            // Fail
+            measurement_url(nil);
+            NSLog(@"error : %@", error.description);
+        }
```
bassosimone: Please, refactor this code as:

```ObjectiveC
if (error) {
  // handle error
  return
}
// handle normal case
```

There is a bunch of reasons for doing so. Not least that most people expect code to be like that.

ooniprobe/Model/Database/Measurement.mm
```diff
@@ -96,4 +96,24 @@ -(void)deleteObject{
     [self remove];
 }
 
+- (void)getExplorerUrl:(void (^)(NSString *))measurement_url {
+    NSString *path = [NSString stringWithFormat:@"https://api.ooni.io/api/v1/measurements?report_id=%@&input=%@", self.report_id, self.url_id.url];
+    NSURL *url = [NSURL URLWithString:path];
+    NSURLSessionDataTask *downloadTask = [[NSURLSession sharedSession] dataTaskWithURL:url completionHandler:^(NSData *data, NSURLResponse *response, NSError *error) {
+        if (!error) {
+            NSDictionary *dic = [NSJSONSerialization JSONObjectWithData:data options:0 error:&error];
```
bassosimone: Please, after you refactor, make sure you handle an error when loading the JSON.

ooniprobe/Model/Database/Measurement.mm
```diff
@@ -96,4 +96,24 @@ -(void)deleteObject{
...
+            NSArray *resultsArray = [dic objectForKey:@"results"];
+            if ([resultsArray count] == 0)
+                measurement_url(nil);
+            measurement_url([[resultsArray objectAtIndex:0] objectForKey:@"measurement_url"]);
```
bassosimone: The above code seem wrong to me. In the case in which `resultsArray` is empty you end up calling `measurement_url` twice. The  second time, you're going out of bounds.

ooniprobe/Model/Database/Measurement.mm
```diff
@@ -96,4 +96,24 @@ -(void)deleteObject{
     [self remove];
 }
 
+- (void)getExplorerUrl:(void (^)(NSString *))measurement_url {
```
bassosimone: Please:

1. refactor the callback to take two arguments, an error and a string, and pass both of them to the callback (this is much better than using a `nil` string as a sentinel for indicating error)

2. rename the callback to be `callback` (or, `cb`, or, if you don't like this naming, which however is quite well established in some arenas, call the callback something like `onMeasurementURL`)

bassosimone:
> I just don't understand why we need the URL.
> Can't we call this function for non-webconnectivity tests?

I do not understand your question. However, now that I think about it, it's quite likely you need to write more flexible code, because there's no input for e.g. `ndt`.

ooniprobe/Model/Database/Measurement.mm
lorenzoPrimi: measurement_url is a completion function and can be called only once if I read the documentation correctly. But I can refactor it to call it once

ooniprobe/Model/Database/Measurement.mm
lorenzoPrimi: 1. The errors will be handled directly in the function, no need to be handled outside it. Much more because this function will be called from different places.

2. I don't understand your suggestion

lorenzoPrimi:
> > I just don't understand why we need the URL.
> > Can't we call this function for non-webconnectivity tests?
> 
> I do not understand your question. However, now that I think about it, it's quite likely you need to write more flexible code, because there's no input for e.g. `ndt`.

Exactly, I didn't understand in the thread https://github.com/ooni/probe/issues/810 why input is a compulsory parameter

ooniprobe/Model/Database/Measurement.mm
lorenzoPrimi: For what I know calling the callback is a return

ooniprobe/Model/Database/Measurement.mm
bassosimone: > [...] and can be called only once if I read the documentation correctly. But I can refactor it to call it once

Yes, this is a common pattern for callback (or completion) functions. In practice this means that adding a return inside the `if` branch is the expected behaviour.

ooniprobe/Model/Database/Measurement.mm
bassosimone: 1. passing `nil` to indicate failure is an anti-pattern, please don't do that!

2. I mean that `measurement_url` is not a good name for a callback.

lorenzoPrimi: This function returns the `measurement_url`. If the measurement url doesn't exists return `nil`. This doesn't indicate a failure, I just care if the measurement url exists or not because I have to delete from db the json with a measurement urls

bassosimone:
> Exactly, I didn't understand in the thread ooni/probe#810 why input is a compulsory parameter

Well, actually, the only example provided there is for a `ndt` test and the URL is without input. So, I guess you can rest assured that `input` is required only when a measurement takes input.

ooniprobe/Model/Database/Measurement.h
```diff
@@ -38,5 +38,5 @@
 -(void)save;
 -(void)setReRun;
 -(void)deleteObject;
--(void)getExplorerUrl:(void (^)(NSString *))measurement_url;
+-(void)getExplorerUrlSuccessCallback:(void (^)(NSString*))success errorCallback:(void (^)(NSError*))error;
```
bassosimone: I find the name `getExplorerIUrlSuccessCallback` a bit verbose. How about (considering usage):

```ObjectiveC
[measurement getExplorerUrl:^(NSString *url) {
  // success case
} otherwise:^(NSError *error) {
  // failure case
}];
```

?

ooniprobe/Model/Database/Measurement.mm
```diff
@@ -96,22 +96,30 @@ -(void)deleteObject{
     [self remove];
 }
 
-- (void)getExplorerUrl:(void (^)(NSString *))measurement_url {
-    NSString *path = [NSString stringWithFormat:@"https://api.ooni.io/api/v1/measurements?report_id=%@&input=%@", self.report_id, self.url_id.url];
+-(void)getExplorerUrlSuccessCallback:(void (^)(NSString*))success errorCallback:(void (^)(NSError*))error {
```
bassosimone: Nit: I think `error` is an instance on error. Better calling the variable `errorcb` (and `successcb` the success callback.)

ooniprobe/Model/Database/Measurement.mm
```diff
@@ -96,22 +96,30 @@ -(void)deleteObject{
...
+    NSLog(@"%@ getExplorerUrl",self.Id);
+    NSMutableString *path = [NSMutableString stringWithFormat:@"https://api.ooni.io/api/v1/measurements?report_id=%@",
+                      self.report_id];
+    if ([self.test_name isEqualToString:@"web_connectivity"])
+        [path appendString:[NSString stringWithFormat:@"&input=%@",
+                            self.url_id.url]];
```
bassosimone: I would suggest here to see whether there is a Cocoa facility for building URLs in a robust way. I am especially concerned with proper escaping. I bet there is a safe Cocoa way to construct a query string and then attaching it to a URL in a way that escaping is on the library rather than on us.

ooniprobe/Model/Database/Measurement.mm
```diff
@@ -96,4 +96,32 @@ -(void)deleteObject{
...
+    NSURL *url = [NSURL URLWithString:path];
+    NSURLSessionDataTask *downloadTask =
+    [[NSURLSession sharedSession]
+     dataTaskWithURL:url
+     completionHandler:^(NSData *data, NSURLResponse *response, NSError *nserror) {
+         if (nserror != nil)
+             error(nserror);
```
bassosimone: Please, don't increase the code complexity. Just return after an error.

ooniprobe/Model/Database/Measurement.mm
```diff
@@ -96,4 +96,32 @@ -(void)deleteObject{
...
+         else {
+             NSDictionary *dic = [NSJSONSerialization JSONObjectWithData:data options:0 error:&nserror];
+             if (nserror != nil)
+                 error(nserror);
```
bassosimone: Same here, just return after an error.

ooniprobe/Model/Database/Measurement.mm
```diff
@@ -96,4 +96,32 @@ -(void)deleteObject{
...
+             NSArray *resultsArray = [dic objectForKey:@"results"];
+             //TODO symbolize somehow empty array
+             if ([resultsArray count] == 0)
+                 error(nil);
```
bassosimone: This code continues to be broken. One subtle detail with two callbacks (they approach that you choose) is that you should call either one or the other but never both. (With a single callback rules are simpler.)

ooniprobe/Model/Database/Measurement.mm
```diff
@@ -96,4 +96,32 @@ -(void)deleteObject{
...
+    NSMutableString *path = [NSMutableString stringWithFormat:@"https://api.ooni.io/api/v1/measurements?report_id=%@",
+                      self.report_id];
+    if ([self.test_name isEqualToString:@"web_connectivity"])
+        [path appendString:[NSString stringWithFormat:@"&input=%@",
+                            self.url_id.url]];
+    NSURL *url = [NSURL URLWithString:path];
+    NSURLSessionDataTask *downloadTask =
+    [[NSURLSession sharedSession]
+     dataTaskWithURL:url
+     completionHandler:^(NSData *data, NSURLResponse *response, NSError *nserror) {
```
bassosimone: Conversationally, you pointed out that AFNetworking is using two callbacks. Maybe I'm missing something, but here above I do actually see that the pattern used is the one where there is a single callback.

ooniprobe/Model/Database/Measurement.mm
bassosimone: No, this is not correct. You need to `return` explicitly right after calling the callback. The code inside a block is not special and all the usual code flow rules apply.

ooniprobe/Model/Database/Measurement.mm
bassosimone: _Errata corrige_: this is NSURLSession, which is Apple code, rather than AFNetworking.

ooniprobe/Model/Database/Measurement.mm
lorenzoPrimi: I don't understand what you say.
Didn't you say we need to call the error callback with the NSError object?

ooniprobe/Model/Database/Measurement.mm
lorenzoPrimi: Should we just integrate AFNetworking into the project that is a more robust library?

ooniprobe/Model/Database/Measurement.mm
lorenzoPrimi: Same comment as above: if there is an error we need to notify the error callback, if everything is fine we trigger the success callback. 
Both can't be triggered, as long as one is triggered the function returns.

lorenzoPrimi: My bad, just checked the docs and the function continues even after the cbs, weird implementation, I would have thought the opposite.
Added the returns but at this point I would prefer to use the AFNetworking class for semplicity

ooniprobe/Model/Database/Measurement.mm
```diff
@@ -96,4 +96,35 @@ -(void)deleteObject{
...
+             if (error != nil)
+                 errorcb(error);
+             NSArray *resultsArray = [dic objectForKey:@"results"];
+             //TODO symbolize somehow empty array
+             if ([resultsArray count] > 0)
+                 errorcb([NSError errorWithDomain:@"io.ooni.api"
+                                           code:100
+                                       userInfo:@{NSLocalizedDescriptionKey:@"Error.JsonEmpty"
+                                                  }]);
+             successcb([[resultsArray objectAtIndex:0] objectForKey:@"measurement_url"]);
```
bassosimone: This code is still not working as intended. Please, apply this pattern:

```ObjectiveC
{
  completionHandler:^(NSData *data, NSURLResponse *response, NSError *error) {
    if (error != nil) {
      errorcb(error);
      return;
    }
    NSDictionary *dic = [NSJSONSerialization JSONObjectWithData:data options:0 error:&error];
    if (error != nil) {
      errorcb(error);
      return;
    }
    NSArray *resultsArray = [dic objectForKey:@"results"];
    //TODO symbolize somehow empty array
    if ([resultsArray count] <= 0) {
        errorcb([NSError errorWithDomain:@"io.ooni.api"
                                                          code:100
                                                     userInfo:@{NSLocalizedDescriptionKey:@"Error.JsonEmpty"
        }]);
        return;
    }
    successcb([[resultsArray objectAtIndex:0] objectForKey:@"measurement_url"]);
}
```

Also, the check on the array count seems to be inverted.

bassosimone: This pattern is better because the code does not grow out of control because of indenting and you can have several clauses. Also, this is how many programmers expect to see the code. This is also for historical reasons as the compilers often assumes that jumps are not taken. Therefore, historically, writing code like this was more performant (and may even be today, but that's really not the point here). Also, and this is actually much more useful to keep in mind, if you organise code like this, you can understand what happens in the "all good path" by just avoid reading the `if`s and their body.

ooniprobe/Model/Database/Measurement.h
```diff
@@ -38,5 +38,5 @@
 -(void)save;
 -(void)setReRun;
 -(void)deleteObject;
-
+-(void)getExplorerUrl:(void (^)(NSString*))successcb error:(void (^)(NSError*))errorcb;
```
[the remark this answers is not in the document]
lorenzoPrimi: What is your suggestion?

ooniprobe/Model/Database/Measurement.mm
```diff
@@ -96,4 +96,39 @@ -(void)deleteObject{
...
+         NSDictionary *dic = [NSJSONSerialization JSONObjectWithData:data options:0 error:&error];
+         if (error != nil) {
+             errorcb(error);
+             return;
+         }
+         NSArray *resultsArray = [dic objectForKey:@"results"];
+         //TODO symbolize somehow empty array
+         if ([resultsArray count] <= 0) {
+             errorcb([NSError errorWithDomain:@"io.ooni.api"
+                                         code:100
```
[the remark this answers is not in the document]
lorenzoPrimi: Give me a code

ooniprobe/Model/Database/Measurement.mm
bassosimone: > [...] weird implementation [...]

Why do you find it weird? It seems consistent to me that code inside a block behaves exactly as code outside the block. I'd have found weird the opposite, TBH.

ooniprobe/Model/Database/Measurement.mm
bassosimone: I am missing data points useful to properly reply to that. My tentative response is that we're probably okay with not adding extra libraries, unless NSURL is demonstrably broken, mainly because we don't want extra dependencies for maintenance and code size reasons.

ooniprobe/Utility/TestUtility.mm
```diff
@@ -72,7 +72,7 @@ + (UIColor*)getColorForTest:(NSString*)testName alpha:(CGFloat)alpha{
     return [UIColor colorWithRGBHexString:color_blue5 alpha:alpha];
 }
 
-+ (void)downloadUrls:(void (^)(NSArray *))completion {
++ (void)downloadUrls:(void (^)(NSArray*))successcb onError:(void (^)(NSError*))errorcb {
```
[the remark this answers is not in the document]
lorenzoPrimi: yes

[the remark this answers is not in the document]
bassosimone: Yeah, you actually did. I was confused by the name `ooniprobe/Test/Test/WebConnectivity.mm`. I usually use the name `Test` or the `Test` folder for unit and/or integration tests. While here it means network tests.

ooniprobe/Model/Database/Measurement.mm
```diff
@@ -96,4 +96,38 @@ -(void)deleteObject{
...
+             errorcb(error);
+             return;
+         }
+         NSDictionary *dic = [NSJSONSerialization JSONObjectWithData:data options:0 error:&error];
+         if (error != nil) {
+             errorcb(error);
+             return;
+         }
+         NSArray *resultsArray = [dic objectForKey:@"results"];
+         //TODO symbolize somehow empty array
```
bassosimone: Is this comment still current? What does it mean?

ooniprobe/Model/Database/Measurement.mm
```diff
@@ -96,4 +96,38 @@ -(void)deleteObject{
...
+         if ([resultsArray count] <= 0) {
+             errorcb([NSError errorWithDomain:@"io.ooni.api"
+                                         code:ERR_JSON_EMPTY
+                                     userInfo:@{NSLocalizedDescriptionKey:@"Error.JsonEmpty"
+                                                }]);
+             return;
+         }
+         successcb([[resultsArray objectAtIndex:0] objectForKey:@"measurement_url"]);
```
bassosimone: I believe we should also check for the case where the `measurement_url` key does not exist.

ooniprobe/Utility/TestUtility.mm
```diff
@@ -86,24 +86,31 @@ + (void)downloadUrls:(void (^)(NSArray *))completion {
...
+         for (NSDictionary* current in urlsArray){
+             //List for database
+             Url *url = [Url checkExistingUrl:[current objectForKey:@"url"] categoryCode:[current objectForKey:@"category_code"] countryCode:[current objectForKey:@"country_code"]];
+             //List for mk
+             [urls addObject:url.url];
+         }
+         if ([urls count] == 0){
+             errorcb(error);
+             return;
+         }
```
bassosimone: This check should be moved before looping over the array.

ooniprobe/Model/Database/Measurement.mm
bassosimone: Do we agree that we'll reconsider AFNetworking later?

lorenzoPrimi: sure, better later

ooniprobe/Model/Database/Measurement.mm
lorenzoPrimi: Nope, it was only to remind me to do some callback if the array is empty. done now

ooniprobe/Utility/TestUtility.mm
lorenzoPrimi: It's correct like it is, review the code, `urls` is an array I initialize in the loop

ooniprobe/Model/Database/Measurement.mm
lorenzoPrimi: done

ooniprobe/Model/Database/Measurement.mm
```diff
@@ -96,4 +96,41 @@ -(void)deleteObject{
...
+    NSURLComponents *components = [[NSURLComponents alloc] init];
+    components.scheme = @"https";
+    components.host = @"api.ooni.io";
+    components.path = @"/api/v1/measurements";
+    if ([self.test_name isEqualToString:@"web_connectivity"])
+        components.query = [NSString stringWithFormat:@"report_id=%@&input=%@",
+                            self.report_id, self.url_id.url];
+    else
+        components.query = [NSString stringWithFormat:@"report_id=%@",
+                        self.report_id];
```
bassosimone: Please, use the NSURLQueryItem API to fill the query items. You can read https://content.pivotal.io/blog/building-urls-with-nsurlqueryitems-and-nsurlcomponents to see how this could be done in practice.

ooniprobe/Model/Database/Measurement.mm
```diff
@@ -96,4 +96,41 @@ -(void)deleteObject{
...
+    NSURL *url = components.URL;
+    NSURLSessionDataTask *downloadTask = [[NSURLSession sharedSession]
+     dataTaskWithURL:url
```
bassosimone: Can you indent the continuation code using two spaces rather than one space? (Or four spaces if you fancy that more, I'm mostly concerned that I see a one-space-indentation unusual)

ooniprobe/Model/Database/Measurement.mm
bassosimone: I believe the error is changed and it's not necessarily that the JSON is empty. Maybe we should use a better error naming and characterisation such as, e.g., "Error.JsonMalformed"?

ooniprobe/Utility/TestUtility.mm
bassosimone: You are right. Still, in this case we're passing to the caller a `nil` error . You should generate an error for this case like you are doing for the above code when you create a `NSError`.

ooniprobe/Utility/TestUtility.mm
```diff
@@ -86,24 +86,31 @@ + (void)downloadUrls:(void (^)(NSArray *))completion {
...
-                [urls addObject:url.url];
-            }
-            completion(urls);
-        }
-        else {
-            // Fail
-            completion(nil);
-            NSLog(@"error : %@", error.description);
-        }
+    NSURLSessionDataTask *downloadTask = [[NSURLSession sharedSession]
```
bassosimone: Since you modified this code, can you also write tests for that (unless they're already present)?

ooniprobe/Model/Database/Measurement.mm
lorenzoPrimi: I can add random spaces but this is how the XCode auto-indentation indented this line

bassosimone: That surprises me, however let's keep nonetheless it like XCode indents it by default.

ooniprobe/Utility/TestUtility.mm
lorenzoPrimi: this class also needs to be modified by adding `NSURLComponents` either we do everything in this PR or in another PR

bassosimone: Okay, since we started changing this code in this PR, let's also add `NSURLComponents` here.

ooniprobe/Utility/TestUtility.mm
```diff
@@ -86,24 +86,34 @@ + (void)downloadUrls:(void (^)(NSArray *))completion {
...
+             //List for database
+             Url *url = [Url checkExistingUrl:[current objectForKey:@"url"] categoryCode:[current objectForKey:@"category_code"] countryCode:[current objectForKey:@"country_code"]];
+             //List for mk
+             [urls addObject:url.url];
+         }
+         if ([urls count] == 0){
+             errorcb([NSError errorWithDomain:@"io.ooni.api"
+                                         code:ERR_JSON_EMPTY
+                                     userInfo:@{NSLocalizedDescriptionKey:@"Error.JsonEmpty"
+                                                }]);
```
bassosimone: Is it possible to factor the duplicated code used to generate this error? Also, I think it's not correct to say that the error here is an empty JSON. It's a JSON that does not contain what we expect.

lorenzoPrimi: Actually it's exactly an empty JSON without urls
`[urls count] == 0`

bassosimone: That's not correct. A JSON like `{"foo": "bar"}` is not empty but will cause the code to report to the caller than the error is `ERR_JSON_EMPTY`, where instead it's "error JSON structure different from expected".

bassosimone: Now that I see the code, I think you should probably also validate that the fields like `"category_code"` are not empty and are indeed strings.

OONIProbeUnitTests/APITest.m
```diff
@@ -0,0 +1,49 @@
+#import <XCTest/XCTest.h>
```
bassosimone: This is just a comment on the directory name. We cannot call this test a "unit test" because it's clearly using the network and relying on its availability. Why are tests folders named like this? Do we want to have a clear separation between unit, integration, and UI tests? Then, probably, we need a folder for integration tests.

OONIProbeUnitTests/APITest.m
```diff
@@ -0,0 +1,49 @@
...
+    measurement.report_id = NONEXISTING_REPORT_ID;
+    [measurement getExplorerUrl:^(NSString *measurement_url){
+        XCTAssert(false);
+    } onError:^(NSError *error){
+        XCTAssert(true);
+    }];
+}
+
+-(void)testDownloadUrls{
+    [TestUtility downloadUrls:^(NSArray *urls) {
```
bassosimone: Please, create different test classes. Create one for `Measurement` and one for `TestUtility`.

ooniprobe/Model/Database/Measurement.mm
```diff
@@ -96,4 +96,44 @@ -(void)deleteObject{
     [self remove];
 }
 
+-(void)getExplorerUrl:(void (^)(NSString*))successcb onError:(void (^)(NSError*))errorcb {
+    NSURLComponents *components = [[NSURLComponents alloc] init];
+    components.scheme = @"https";
+    components.host = @"api.ooni.io";
+    components.path = @"/api/v1/measurements";
+    NSURLQueryItem *reportIdItem = [NSURLQueryItem queryItemWithName:@"report_id" value:self.report_id];
```
bassosimone: What gives us confidence that `report_id` and `url_id` are not `nil`?

ooniprobe/Utility/TestUtility.mm
```diff
@@ -72,38 +72,56 @@ + (UIColor*)getColorForTest:(NSString*)testName alpha:(CGFloat)alpha{
...
+    NSURLComponents *components = [[NSURLComponents alloc] init];
+    components.scheme = @"https";
+    components.host = @"orchestrate.ooni.io";
+    components.path = @"/api/v1/test-list/urls";
+    NSURLQueryItem *ccItem = [NSURLQueryItem queryItemWithName:@"country_code" value:cc];
     if ([[SettingsUtility getSitesCategoriesDisabled] count] > 0){
         NSMutableArray *categories = [NSMutableArray arrayWithArray:[SettingsUtility getSitesCategories]];
         [categories removeObjectsInArray:[SettingsUtility getSitesCategoriesDisabled]];
-        path = [NSString stringWithFormat:@"%@&category_codes=%@", path, [categories componentsJoinedByString:@","]];
+        NSURLQueryItem *categoriesItem = [NSURLQueryItem queryItemWithName:@"category_codes" value:[categories componentsJoinedByString:@","]];
```
bassosimone: Please, wrap this line, as it's too long

ooniprobe/Utility/TestUtility.mm
bassosimone: And also: 

> Is it possible to factor the duplicated code used to generate this error?

ooniprobe/Utility/TestUtility.mm
```diff
@@ -72,38 +72,56 @@ + (UIColor*)getColorForTest:(NSString*)testName alpha:(CGFloat)alpha{
...
+         NSDictionary *dic = [NSJSONSerialization JSONObjectWithData:data options:0 error:&error];
+         if (error != nil) {
+             errorcb(error);
+             return;
+         }
+         NSArray *urlsArray = [dic objectForKey:@"results"];
+         NSMutableArray *urls = [[NSMutableArray alloc] init];
+         for (NSDictionary* current in urlsArray){
+             //List for database
+             Url *url = [Url checkExistingUrl:[current objectForKey:@"url"] categoryCode:[current objectForKey:@"category_code"] countryCode:[current objectForKey:@"country_code"]];
```
bassosimone: Please, wrap this long line

OONIProbeUnitTests/APITest.m
lorenzoPrimi: The dir name is the target name, to change this we should create a new target and this increase complexity. Maybe these tests didn't belong in this PR

OONIProbeUnitTests/APITest.m
lorenzoPrimi: These are API tests, they belong together. The function in `Measurement` could have been written in the TestUtility class (like I did in the first commit) or vice versa. Eventually the `TestUtility` class has to be refactored

ooniprobe/Model/Database/Measurement.mm
lorenzoPrimi: Url can't be due to the db structure. We never call this function in non upload measurements, so report_id will alwayd be there.
To answer your question: if they are `nil` there will simply be a nil in the url, and the function will return non existent measurement_url

ooniprobe/Utility/TestUtility.mm
lorenzoPrimi: so this function requires heavy refactor:
- check url, category_code and country_code not nil, in case they are discard the input
- change ERR_JSON_EMPTY to ERR_NO_VALID_URLS or something similar, meaning that every url sent by the api is invalid

OONIProbeUnitTests/APITest.m
bassosimone: > Maybe these tests didn't belong in this PR

Tests are always welcome.

You didn't reply to my question about why test names have been chosen like this. What was the rationale?

OONIProbeUnitTests/APITest.m
bassosimone: I really do not understand all this resistance to making things right. Also:

> X is fine [...] because eventually will be refactored

is the biggest lie in the history of software engineering. Why not trying to do things right _now_?

ooniprobe/Utility/TestUtility.mm
bassosimone: [sudo please wrap this long line](https://xkcd.com/149/)?

ooniprobe/Utility/TestUtility.mm
bassosimone: My suggestion is to do a little bit of fake [TDD](https://it.wikipedia.org/wiki/Test_driven_development) here. We do refactor the code such that we have a function taking in input the `NSData *data`. The callback block will check whether there were errors and then call such new function. Then we write tests in which we pass data to the new function such that we provoke all the possible error conditions that we can see. We will probably see the code fail, we will learn how it is failing and decide how to repair.

bassosimone: Also, `ERR_JSON_EMPTY` is probably not a description of the right error we see here.

OONIProbeUnitTests/APITest.m
lorenzoPrimi: Because it's not a folder, it's a target/scheme:
Arturo choose the second
Xcode choose the third
Screenshot 2019-06-27 at 16 38 53

OONIProbeUnitTests/APITest.m
lorenzoPrimi: What I'm saying is that it makes sense to keep all the API related calls in one class rather than around many classes. And in this case we'll need just one test.

ooniprobe/Utility/TestUtility.mm
lorenzoPrimi: This line has to be removed as there is need to check for url country_code and category_code
https://github.com/ooni/probe-ios/pull/293#discussion_r298369642

OONIProbeUnitTests/APITest.m
bassosimone: Right, I guess I need to learn what is a target/scheme. TBH it also looks like a folder when seen from GitHub and from the filesystem, so I think my ignorance can be partly excused.

bassosimone: So, each type of tests is a different target? There is a scheme for each target? Is this understanding correct? Can you also clarify the "it's not a folder" part?

OONIProbeUnitTests/APITest.m
bassosimone: > it makes sense to keep all the API related calls in one class

Why? Is this an iOS pattern I'm not aware of?

My instinct here suggests rather the opposite. When I am looking for tests for `.../X.m` I usually search for the corresponding `...Tests/XTest.m` file.

lorenzoPrimi: Well it's a nice patterns, as the `TestUtility` class was meant to be for something else. 
And was not coded at all in Android as xan disliked that pattern at all.
Now it's collecting a lot of functions "I don't know where this belongs"

OONIProbeUnitTests/APITest.m
lorenzoPrimi: Xcode by default create a folder for every target, if you create OONI Probe Watch for example will create a folder with that name

lorenzoPrimi: See different test targets here
Screenshot 2019-06-27 at 16 52 31

ooniprobe/Model/Database/Measurement.mm
bassosimone: > Url can't be due to the db structure. We never call this function in non upload measurements, so report_id will alwayd be there.

Thanks for the explanation. This is true _now_ but may change in the future. Especially with implicitly arguments that are class members. 🤔 

> To answer your question: if they are nil there will simply be a nil in the url, and the function will return non existent measurement_url

Right. I guess in such case the app would misbehave and we'd not clearly know why. I am quite sad that we're in a language that allows `nil`, but I guess this is it.

Are `report_id` and `url_id` externally settable? Maybe it makes sense to fail the test with a specific error in case they are not correctly set (i.e. they're `null`)?

ooniprobe/Utility/TestUtility.mm
bassosimone: Gotcha!

OONIProbeUnitTests/APITest.m
bassosimone: I see. Then, we cannot easily rename now. It's fine to put there also non-unit tests. We'll rename later. :-)

OONIProbeUnitTests/APITest.m
bassosimone: I understand. Can you add a `// TODO(lorenzoPrimi): ...` comment indicating that we should perform this kind of refactor, so that a reader stumbling upon that is not surprised?

ooniprobe/Model/Database/Measurement.mm
```diff
@@ -96,4 +96,48 @@ -(void)deleteObject{
...
+-(void)getExplorerUrl:(void (^)(NSString*))successcb onError:(void (^)(NSError*))errorcb {
+    NSURLComponents *components = [[NSURLComponents alloc] init];
+    components.scheme = @"https";
+    components.host = @"api.ooni.io";
+    components.path = @"/api/v1/measurements";
+    NSURLQueryItem *reportIdItem = [NSURLQueryItem
+                                    queryItemWithName:@"report_id"
+                                    value:self.report_id];
+
+    if ([self.test_name isEqualToString:@"web_connectivity"]){
```
bassosimone: Maybe here a comment should explain that `web_connectivity` is the only test using `input` for now?

OONIProbeUnitTests/APITest.m
bassosimone: Yeah, I guess there's no immediate action for now.

ooniprobe/Model/Database/Measurement.mm
lorenzoPrimi: `report_id` is set only by the value returned by mk
`url_id` is an `Url` db object and must have a string value, so it can't be nil

ooniprobe/Utility/TestUtility.mm
lorenzoPrimi: I want to avoid dealing with `NSData` as much as possible. But maybe I didn't understand what you want to code here. The `NSData` must be managed in the `NSURLSessionDataTask` function so it can call the relative callbacks

bassosimone: I mean the following:

```ObjectiveC
+ (void)downloadUrlsCallback:(NSData *)data
                    response:(NSURLResponse *)response
                       error:(NSError *)error)
                   onSuccess:(void (^)(NSArray*))successcb
                     onError:(void (^)(NSError*))errorcb {
  // qua ci metti il codice della callback originale
}

+ (void)downloadUrls:(void (^)(NSArray*))successcb onError:(void (^)(NSError*))errorcb {
  // qua tutto come prima tranne questo
  NSURLSessionDataTask *downloadTask = [[NSURLSession sharedSession]
   dataTaskWithURL:url
   completionHandler:^(NSData *data, NSURLResponse *response, NSError *error) {
     [TestUtility downloadUrlsCallback:data response:response error:error
      onSuccess:successcb onError:errorcb];
   }]];
  [downloadTask resume];
}
```

Then tests can call `downloadUrlsCallback` with specific `NSData *` that cause additional errors and we are sure that we cover all the relevant code paths.

ooniprobe/Utility/TestUtility.mm
```diff
@@ -72,38 +72,61 @@ + (UIColor*)getColorForTest:(NSString*)testName alpha:(CGFloat)alpha{
...
```
bassosimone: Can you wrap this long line?

ooniprobe/Utility/TestUtility.mm
```diff
@@ -72,38 +72,61 @@ + (UIColor*)getColorForTest:(NSString*)testName alpha:(CGFloat)alpha{
...
+             return;
+         }
+         NSDictionary *dic = [NSJSONSerialization JSONObjectWithData:data options:0 error:&error];
+         if (error != nil) {
+             errorcb(error);
+             return;
+         }
+         NSArray *urlsArray = [dic objectForKey:@"results"];
+         NSMutableArray *urls = [[NSMutableArray alloc] init];
+         for (NSDictionary* current in urlsArray){
```
bassosimone: Will the loop work if the `urlsArray` is `nil`? Or will it crash?

ooniprobe/Utility/TestUtility.mm
lorenzoPrimi: Also added nil checks

ooniprobe/Utility/TestUtility.mm
lorenzoPrimi: If it's nil it works like an empty array: doesn't go into the loop

ooniprobe/Model/Database/Measurement.mm
```diff
@@ -96,4 +96,57 @@ -(void)deleteObject{
...
+    }
+    NSArray *resultsArray = [dic objectForKey:@"results"];
+    if ([resultsArray count] <= 0 || ![[resultsArray objectAtIndex:0] objectForKey:@"measurement_url"]) {
+        errorcb([NSError errorWithDomain:@"io.ooni.api"
+                                    code:ERR_JSON_EMPTY
+                                userInfo:@{NSLocalizedDescriptionKey:@"Error.JsonEmpty"
+                                           }]);
+        return;
+    }
+    successcb([[resultsArray objectAtIndex:0] objectForKey:@"measurement_url"]);
```
hellais: You should also be checking if the array is longer than 1. If it's longer than 1 you should probably not delete the measurement file on disk because it's not possible currently to reliably retrieve it.

ooniprobe/Model/Database/Measurement.mm
```diff
@@ -96,4 +96,57 @@ -(void)deleteObject{
...
+        errorcb(error);
+        return;
+    }
+    NSDictionary *dic = [NSJSONSerialization JSONObjectWithData:data options:0 error:&error];
+    if (error != nil) {
+        errorcb(error);
+        return;
+    }
+    NSArray *resultsArray = [dic objectForKey:@"results"];
+    if ([resultsArray count] <= 0 || ![[resultsArray objectAtIndex:0] objectForKey:@"measurement_url"]) {
```
hellais: Why is this not just `== 0`?

bassosimone: It’s a habit that I have: to exclude all cases to which we are not interested even if the type is unsigned, mainly because that is more robust to refactoring.

bassosimone: It’s a habit that I have: to exclude all cases to which we are not interested even if the type is unsigned, mainly because that is more robust to refactoring.

lorenzoPrimi:
For the records, in the specs we said we want this function to be called max 10 concurrent times,
however `sharedSession` already have a max set to 6 concurrent requests, so we'll keep that instead of writing our own queue: https://stackoverflow.com/questions/56728301/max-concurrent-instances-of-a-nsurlsessiondatatask/56730705

PR #295, opened 2019-07-02
Delete json already in the explorer

lorenzoPrimi:


ooniprobe/Utility/TestUtility.mm
```diff
@@ -144,6 +144,14 @@ + (void)downloadUrlsCallback:(NSData *)data
     successcb(urls);
 }
 
+//TODO where and when to call it?
++ (void)deleteUploadedJsons{
+    for (Measurement *measurement in [Measurement measurementsWithJson]){
+        [measurement getExplorerUrl:^(NSString *measurement_url){
+            [TestUtility removeFile:[measurement getReportFile]];
+        } onError:nil];
```
bassosimone: Can you write a test where you fill in, say, two measurements that are in explorer and one that is not, and verify that the files that should be removed are removed and the other is still there? You can operate like this:

1. save the measurement JSON from explorer or the API (I can help with that)

2. create the files and populate the `Measurement`

3. call `deleteUploadedJsons` and then verify that what happened is what we expected

bassosimone:
(Nevermind about the history, when squashing we'll just get the diff between this branch and master committed to master, so it's really no big deal.)

ooniprobe/Utility/TestUtility.mm
```diff
@@ -144,7 +144,23 @@ + (void)downloadUrlsCallback:(NSData *)data
...
++ (void)deleteUploadedJsons:(void (^)(Measurement*))successcb onError:(void (^)(NSError*))errorcb {
+    for (Measurement *measurement in [Measurement measurementsWithJson]){
+        [measurement getExplorerUrl:^(NSString *measurement_url){
+            [TestUtility removeFile:[measurement getReportFile]];
+            successcb(measurement);
+        } onError:^(NSError *error){
+            errorcb(error);
+        }];
+    }
+}
```
bassosimone: So, here's an alternative idea. Let me know what are your thoughts:

```ObjectiveC

+ (void)deleteUploadedJsonsWithMeasurementRemover:(void (^)(Measurement *))remover {
    for (Measurement *measurement in [Measurement measurementsWithJson]) {
        [measurement getExplorerUrl:^(NSString *measurement_url){
            remover(measurement);
        } onError:nil];
    }
}

+ (void)deleteUploadedJsons{
    [self deleteUploadedJsonsWithMeasurementRemover:^(NSString *path) {
        [TestUtility removeFile:[measurement getReportFile]];
    }];
}
```

Then:

1. in the callback in the tests you assert that you only see the measurements you expected to see

2. you use the expectation pattern with a `60` seconds timeout and declare the test failed if we are not able to walk through the measurements in such timeout

This way you don't need to be concerned about counting and you don't need to be concerned about knowing when you have walked through all the list. You only know it's reasonable for it to finish in `60` seconds.

(It's probably a bit of a hack, but my current level of ObjectiveC-fu is such that I cannot think of anything better that does not also make code more complex. If we had coroutines, this would be simpler.)

ooniprobe/Utility/TestUtility.mm
```diff
@@ -144,7 +144,23 @@ + (void)downloadUrlsCallback:(NSData *)data
...
+
+//TODO where and when to call it?
++ (void)deleteUploadedJsons{
+    [self deleteUploadedJsons:nil onError:nil];
+}
+
++ (BOOL)removeFile:(NSString*)fileName {
```
bassosimone: It seems we're not ever using the return value, so we probably don't need to propagate it? (Or am I missing something?)

OONIProbeUnitTests/APITest.m
```diff
@@ -68,4 +69,39 @@ -(void)testDownloadUrls{
...
+            [expectation fulfill];
+    } onError:^(NSError *error){
+        if ([measurement.report_id isEqualToString:NONEXISTING_REPORT_ID])
+            XCTAssert(true);
+        else
+            XCTAssert(false);
+        i--;
+        if (i == 0)
+            [expectation fulfill];
+    }];
```
bassosimone: Let me know if my proposal below can help to test without making the code too hairy. Also, I believe that there is a singleton for measurements, so upon leaving this test you probably want to reset the database?

ooniprobe/Utility/TestUtility.mm
lorenzoPrimi: I added it in case we wanted to do a unit test only for removeFile

OONIProbeUnitTests/APITest.m
lorenzoPrimi: This code was only a test, it would never worked. I will try your proposal and see if it works

ooniprobe/Utility/TestUtility.mm
lorenzoPrimi: Can you write the test even as pseudo code? 
I don't see what exactly you are suggesting here.
Are you suggesting that we don't actually remove the json, only to be sure that functions return the callback on the two measurements we are expecting to delete?

bassosimone: Yes, I am suggesting that we just check that it the callback is called with the right measurements. This is more a unit test than an integration test, in that we do not verify that files are actually deleted. An integration test may perhaps be written later.

We can also write another test for the function deleting files, maybe not in this specific file tho.

lorenzoPrimi: Did it but I still have some question, check the comment

OONIProbeUnitTests/APITest.m
```diff
@@ -68,4 +69,29 @@ -(void)testDownloadUrls{
...
+    Measurement *existing_1 = [self addMeasurement:EXISTING_REPORT_ID];
+    Measurement *nonexisting_1 = [self addMeasurement:NONEXISTING_REPORT_ID];
+    Measurement *existing_2 = [self addMeasurement:EXISTING_REPORT_ID_2];
+    Measurement *nonexisting_2 = [self addMeasurement:NONEXISTING_REPORT_ID];
+    [self deleteUploadedJsonsWithMeasurementRemover:^(Measurement *measurement) {
+        if ([measurement.report_id isEqualToString:EXISTING_REPORT_ID] || [measurement.report_id isEqualToString:EXISTING_REPORT_ID_2])
+            XCTAssert(true);
+        else
+            XCTAssert(false);
+        //TODO how do i check that this callback has been called for both measurement in the if and fulfill only after that?
```
lorenzoPrimi: @bassosimone

ooniprobe/Utility/TestUtility.mm
bassosimone: Ok

OONIProbeUnitTests/APITest.m
lorenzoPrimi: - You can't edit `failures` and `successess` inside that without passing the vars somehow.
- I don't see any fullfill in this code, where would you put them?
- What the failure value stands for?

bassosimone: How about then using NSInteger rather than just int? Since a NSInteger instance is a pointer to NSInteger, you are able to change its value, right?

I think I forgot to add a fulfill part. In there I think you should assert that you did see two “success” and zero “failure”. Does this make sense?

(It’s still a rather bad test because it runs for ten seconds while you have an answer much early than that...)

lorenzoPrimi: https://stackoverflow.com/questions/7962721/assign-a-variable-inside-a-block-to-a-variable-outside-a-block

lorenzoPrimi: I guess there is no need for the failure var. When it enters the failure part once the test is failed.

lorenzoPrimi: New code pushed, there is still something missing

OONIProbeUnitTests/APITest.m
```diff
@@ -68,4 +70,39 @@ -(void)testDownloadUrls{
...
+                if (successes == 2){
+                    XCTAssert(true);
+                    [expectation fulfill];
+                }
+            }
+        } else {
+            XCTAssert(false);
+            [expectation fulfill];
+        }
+    }];
```
bassosimone: I think the test is still incorrect. You need to wait for the expectation. Additionally, fulfilling the expectation inside of the block is wrong. What you want is to count inside the block and verify when the expectation is fulfilled that you have counted correctly. Lastly, you also need to count the cases where unexpected measurements were called, because you want to make sure you operate only on the measurements where you expected to operate.

lorenzoPrimi:
Travis works!
Only thing left to do is to ensure when running tests there are no files on disk and we start from a zero state.

ooniprobe/Utility/TestUtility.mm
```diff
@@ -144,7 +144,23 @@ + (void)downloadUrlsCallback:(NSData *)data
...
++ (void)deleteUploadedJsonsWithMeasurementRemover:(void (^)(Measurement *))remover {
+    for (Measurement *measurement in [Measurement measurementsWithJson]) {
+        [measurement getExplorerUrl:^(NSString *measurement_url){
+            remover(measurement);
+        } onError:^(NSError *error) {
+        }];
+    }
+}
+
+//TODO where and when to call it?
```
bassosimone: I guess this comment is redundant because we will do that in the next PR

lorenzoPrimi: Actually not, the next PR is to open a link.
We need to decide here where to call this function
1) On app opening (always/once a week/etc...)
2) On app resuming (every time you go from background to foreground)
3) When opening raw data (a local json) always checking if that test has been uploaded

I think 1 and 3.

bassosimone: 👍 thanks, I was missing that bit

PR #296, opened 2019-07-02
Fix async tests

lorenzoPrimi:


bassosimone:
Please change the label of the last test to be @“testDownloadUrls” rather than being @“testNonExisting”. Apart from that, :whale:

PR #300, opened 2019-07-16
Call delete function

lorenzoPrimi:
When even the app is opened and the function was called more than 24h ago

### DIFF
--- a/ooniprobe/Utility/TestUtility.mm
+++ b/ooniprobe/Utility/TestUtility.mm
@@ -156,7 +156,6 @@
     }
 }
 
-//TODO where and when to call it?
 + (void)deleteUploadedJsons{
     [self deleteUploadedJsonsWithMeasurementRemover:^(Measurement *measurement) {
         [TestUtility removeFile:[measurement getReportFile]];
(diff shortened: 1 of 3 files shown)


ooniprobe/Utility/TestUtility.mm
```diff
@@ -154,11 +156,28 @@ + (void)deleteUploadedJsonsWithMeasurementRemover:(void (^)(Measurement *))remov
...
+    [[NSUserDefaults standardUserDefaults] synchronize];
+}
+
++ (BOOL)canCallDeleteJson{
+    NSDate *lastCalled =  (NSDate *)[[NSUserDefaults standardUserDefaults] objectForKey:delete_json_key];
+    
+    if (lastCalled == nil){
+        return YES;
+    }
+    else {
```
bassosimone: This `else` is unnecessary because you `return YES`.

ooniprobe/Utility/TestUtility.mm
```diff
@@ -154,11 +156,28 @@ + (void)deleteUploadedJsonsWithMeasurementRemover:(void (^)(Measurement *))remov
...
+        NSTimeInterval timeOfNoMagic = [[NSDate date] timeIntervalSinceDate:lastCalled];
```
bassosimone: Why is the variable called like this? I don't get it

lorenzoPrimi: Sorry SO copypasta mistake :D

PR #302, opened 2019-07-25
Implement faster measurement resubmission

bassosimone:
Part of https://github.com/ooni/probe/issues/861.

Uses https://github.com/measurement-kit/mkcollector/pull/25 and https://github.com/measurement-kit/mkall-ios/pull/25 to make the act of resubmitting really faster.

This is bundled with https://github.com/measurement-kit/mkall-ios/releases/tag/v0.5.5 that contains also reliability fixes, including a fix to make sure we really deallocate ObjectiveC objects context.

### DIFF
--- a/ooniprobe/View/TestResults/Footer/UploadFooterViewController.m
+++ b/ooniprobe/View/TestResults/Footer/UploadFooterViewController.m
@@ -123,7 +123,6 @@
             reporterTask:(MKReporterTask *)task {
     NSString *content = [TestUtility getUTF8FileContent:[measurement getReportFile]];
     NSUInteger bytes = [content lengthOfBytesUsingEncoding:NSUTF8StringEncoding];
-    //[task setTimeout:[TestUtility makeTimeout:bytes]]; /* XXX */
     MKReporterResults *results = [
       task submit:content uploadTimeout:[TestUtility makeTimeout:bytes]];
     if ([results good]){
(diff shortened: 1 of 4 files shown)


bassosimone:
@lorenzoPrimi please take a look

bassosimone:
@lorenzoPrimi there is a branch and PR of the iOS app that you can use for this purpose

bassosimone:
@lorenzoPrimi updated to use mkall-ios v0.6.0 (and MK v0.10.5). Please, take a look!

ooniprobe/View/TestResults/Footer/UploadFooterViewController.m
```diff
@@ -87,14 +87,17 @@ -(void)uploadMeasurements:(NSArray *)notUploaded startAt:(NSUInteger)idx{
         float progress = 0.0f;
         //The progress should consider the array size - the idx of first element to actually be uploaded
         float measurementValue = 1.0/([notUploaded count] - i);
+        MKReporterTask *task = [[MKReporterTask alloc]
```
lorenzoPrimi: I guess it makes more sense to create `MKReporterTask` inside  `uploadMeasurement` rather than passing it as parameter. Or explain me this implementation.

bassosimone: You want to reuse the same `task` such that you continue using the existing HTTP client, so you reuse already existing connections. Also, the same task keeps the report open as long as needed and then closes the report when you are submitting something completely different. For these two reasons, the `task` needs to have a lifecycle larger than the one of the loop. If we do that inside of `uploadMeasurement`, we lose all the improvements to performance that I mentioned above and we are as small as we used to be.

lorenzoPrimi: Got it thanks!

bassosimone:
As discussed with @lorenzoPrimi, I'll leave to merge this PR when it's most convenient for him!

PR #307, opened 2019-09-04
Checking for internet connection before downloading raw data from explorer

lorenzoPrimi:

### DIFF
--- a/ooniprobe/Notifications/ReachabilityManager.h
+++ b/ooniprobe/Notifications/ReachabilityManager.h
@@ -7,5 +7,5 @@
 @property (strong, nonatomic) Reachability *reachability;
 + (ReachabilityManager *)sharedManager;
 - (NSString*)getStatus;
-- (BOOL)isInternetAccessible;
+- (BOOL)noInternetAccess;
 @end
--- a/ooniprobe/Notifications/ReachabilityManager.h
+++ b/ooniprobe/Notifications/ReachabilityManager.h
@@ -7,4 +7,5 @@
 @property (strong, nonatomic) Reachability *reachability;
 + (ReachabilityManager *)sharedManager;
 - (NSString*)getStatus;
+- (BOOL)isInternetAccessible;
 @end
--- a/ooniprobe/View/TestResults/Details/TestDetailsViewController.m
+++ b/ooniprobe/View/TestResults/Details/TestDetailsViewController.m
@@ -49,7 +49,13 @@
                                    actionWithTitle:NSLocalizedString(@"TestResults.Details.RawData", nil)
                                    style:UIAlertActionStyleDefault
                                    handler:^(UIAlertAction * action) {
-                                       [self rawData];
+                                       if (![[[ReachabilityManager sharedManager] getStatus] isEqualToString:@"no_internet"]
+                                           && isInExplorer)
+                                           [self rawData];
+                                       else
+                                           [MessageUtility
+                                            alertWithTitle:NSLocalizedString(@"Modal.Error", nil)
+                                            message:NSLocalizedString(@"Modal.Error.NoInternet", nil) inView:self];
                                    }];
     UIAlertAction* logButton = [UIAlertAction
                                    actionWithTitle:NSLocalizedString(@"TestResults.Details.ViewLog", nil)
--- a/ooniprobe/View/TestResults/Details/TestDetailsViewController.m
+++ b/ooniprobe/View/TestResults/Details/TestDetailsViewController.m
@@ -50,7 +50,7 @@
                                    actionWithTitle:NSLocalizedString(@"TestResults.Details.RawData", nil)
                                    style:UIAlertActionStyleDefault
                                    handler:^(UIAlertAction * action) {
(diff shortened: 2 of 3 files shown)


ooniprobe/View/TestResults/Details/TestDetailsViewController.m
```diff
@@ -49,7 +49,13 @@ - (void)advancedScreens{
                                    actionWithTitle:NSLocalizedString(@"TestResults.Details.RawData", nil)
                                    style:UIAlertActionStyleDefault
                                    handler:^(UIAlertAction * action) {
-                                       [self rawData];
+                                       if (![[[ReachabilityManager sharedManager] getStatus] isEqualToString:@"no_internet"]
+                                           && isInExplorer)
```
bassosimone: I don't get why `isInExplorer` is checked for here

lorenzoPrimi: Because otherwise we load the text file from disk, so there is no need to check connection

bassosimone: Can you write an utility function to encapsulate the "there is no internet" logic? I was confused.

My understanding is that the logic is like: (1) if we have internet access and the data is in explorer, then return the data (?) otherwise (2) show an error. Do I understand it correctly?

(This seems also a good moment to say again that

```
if (!condition) {
  error();
  return;
}
// ...
```

is generally more readable for me and less complex to parse than structured programming.)

lorenzoPrimi: There is no internet logic is `[[[ReachabilityManager sharedManager] getStatus] isEqualToString:@"no_internet"]`
if you want I can write a feature that incapsulate this code

lorenzoPrimi: we can't do the code format you wrote as we need to execute different code if there is internet or not so I need an if else

lorenzoPrimi: The logic is: if there is no internet and the measurement is in explorer (so it's not in local) show an error.
Else go to the rawData screen that will handle the logic (load from disk or from internet)

bassosimone: > if you want I can write a feature that incapsulate this code

Yes, please!

bassosimone: @nuke what I am saying is that you can write (in pseudo JavaScript):

```javascript
function () {
  if (!haveInternet() || !haveMeasurement) {
    showError();
    return;
  }
  showMeasurement();
}
```

ooniprobe/View/TestResults/Details/TestDetailsViewController.m
```diff
@@ -49,6 +50,13 @@ - (void)advancedScreens{
                                    actionWithTitle:NSLocalizedString(@"TestResults.Details.RawData", nil)
                                    style:UIAlertActionStyleDefault
                                    handler:^(UIAlertAction * action) {
+                                       if (![[ReachabilityManager sharedManager] isInternetAccessible]
```
bassosimone: can you change the helper semantics and call it `noInternetAccess` so we avoid using negations?

lorenzoPrimi: Should I add another function noInternetAccess or remove isInternetAccessible ?

bassosimone: Let's just add what we need

ooniprobe/Notifications/ReachabilityManager.h
```diff
@@ -7,4 +7,5 @@
 @property (strong, nonatomic) Reachability *reachability;
 + (ReachabilityManager *)sharedManager;
 - (NSString*)getStatus;
+- (BOOL)isInternetAccessible;
```
bassosimone: You should also change this line

ooniprobe/View/TestResults/Details/TestDetailsViewController.m
```diff
@@ -49,6 +50,13 @@ - (void)advancedScreens{
                                    actionWithTitle:NSLocalizedString(@"TestResults.Details.RawData", nil)
                                    style:UIAlertActionStyleDefault
                                    handler:^(UIAlertAction * action) {
+                                       if ([[ReachabilityManager sharedManager] isInternetAccessible]
```
bassosimone: you also need to change the name here

ooniprobe/View/TestResults/Details/TestDetailsViewController.m
```diff
@@ -49,6 +50,13 @@ - (void)advancedScreens{
                                    actionWithTitle:NSLocalizedString(@"TestResults.Details.RawData", nil)
```
bassosimone: I think you can delete the `TODO` comment at the beginning of the function

lorenzoPrimi: It has been handled in another branch

PR #315, opened 2019-11-19
Appium testing

lorenzoPrimi:


hellais:
We might want to try out: https://github.com/marketplace/actions/appium-server-action

lorenzoPrimi:
This can be safely closed as there is no edit we need to bing into master

PR #317, opened 2019-11-20
Adding Storyboard IDs

lorenzoPrimi:


lorenzoPrimi:
@bassosimone I only added the ids if there is any other change was made automatically by XCode

lorenzoPrimi:
I think it was just because the device preview changed

PR #321, opened 2019-12-29
Change TextView to WebView for logs and json

lorenzoPrimi:
Fixes https://github.com/ooni/probe-android/issues/229 and https://github.com/ooni/probe/issues/865

## Proposed Changes

  - Changed the `TextView` into `WebView`
  - Using a "text/plain" `WebView`
  - Added a loading wheel for the WebView


ooniprobe/View/TestResults/LogViewController.m
```diff
@@ -20,11 +23,14 @@ - (void)viewDidLoad {
         NSString *fileName = [self.measurement getLogFile];
         NSString *content = [TestUtility getUTF8FileContent:fileName];
         if (content != nil) {
-            [self.textView setText:content];
+            self.text = content;
+            [self.webView loadData:[content dataUsingEncoding:NSUTF8StringEncoding]
```
bassosimone: What happens is the data is not UTF-8? (It should be but that is not enforced on the language level.)

lorenzoPrimi: I guess will encode/show the data in a wrong way

PR #322, opened 2019-12-29
.ruby-version: use ruby 2.7

bassosimone:
This is an attempt at fixing https://travis-ci.org/ooni/probe-ios/builds/630612714


bassosimone:
Explanation: we were using a too old ruby version that depended on a version of openssl that now is not anymore part of the typical setup you have with homebrew.

PR #323, opened 2020-01-03
Allow to enable/disable max_runtime

lorenzoPrimi:
Fixes https://github.com/ooni/probe/issues/893

Missing the string, at the moment it say 0 as runtime


bassosimone:
@lorenzoPrimi I merged `master` into the branch to check whether tests pass.

ooniprobe/Test/Test/WebConnectivity.mm
```diff
@@ -44,15 +45,17 @@ -(void)onEntry:(JsonResult*)json obj:(Measurement*)measurement{
 }
 
 -(int)getRuntime{
-    if (self.settings.options.max_runtime != nil){
+    if (self.settings.options.max_runtime != nil &&
+        ![self.settings.options.max_runtime isEqual: MAX_RUNTIME_DISABLED]){
```
bassosimone: I believe you should check for the runtime being >= 0

ooniprobe/Test/Test/WebConnectivity.mm
```diff
@@ -44,15 +45,17 @@ -(void)onEntry:(JsonResult*)json obj:(Measurement*)measurement{
 }
 
 -(int)getRuntime{
```
bassosimone: I do not understand the logic for setting the `max_runtime`. Namely:

1. I'd check for the setting being disabled at the beginning and return `-1` immediately in such case

2. why is the case where `inputs != nil` different

3. why `self.settings.options.max_runtime` is (I assume) a copy of the settings stored into `WebConnectivity` rather than being just read from the settings

I feel like there is probably a good reason why that is the case, but the lack of documentation is making it more difficult for me to understand what we are doing in this diff.

ooniprobe/View/RunTest/TestOverviewViewController.m
```diff
@@ -46,7 +46,12 @@ -(void)viewWillAppear:(BOOL)animated{
 -(void)reloadLastMeasurement{
     dispatch_async(dispatch_get_main_queue(), ^{
         [self.estimatedLabel setText:NSLocalizedString(@"Dashboard.Overview.Estimated", nil)];
-        NSString *time = NSLocalizedFormatString(@"Dashboard.Card.Seconds", [NSString stringWithFormat:@"%d", [testSuite getRuntime]]);
+        //TODO convert seconds to minutes and hours when needed
+        //if getRuntime = 0 show one hour
```
bassosimone: This comment has now become inaccurate

ooniprobe/View/RunTest/TestOverviewViewController.m
```diff
@@ -46,7 +46,12 @@ -(void)viewWillAppear:(BOOL)animated{
...
+        int runTime = [testSuite getRuntime];
+        if (runTime == [MAX_RUNTIME_DISABLED intValue])
```
bassosimone: I would change the check to say `<=` because that is more robust and the intent is that negative means disabled

ooniprobe/View/RunTest/TestRunningViewController.mm
```diff
@@ -50,9 +51,14 @@ -(void)runTest{
 }
 
 -(void)setRuntime{
-    totalRuntime = [testSuite getRuntime];
-    NSString *time = NSLocalizedFormatString(@"Dashboard.Running.Seconds", [NSString stringWithFormat:@"%d", totalRuntime]);
-    [self.timeLabel setText:time];
+    dispatch_async(dispatch_get_main_queue(), ^{
+        totalRuntime = [testSuite getRuntime];
+        //We don't want to show -1 seconds before downloading the URL list
+        if (totalRuntime == [MAX_RUNTIME_DISABLED intValue])
```
bassosimone: Also here `<=` please?

ooniprobe/View/TestResults/Rows/DashboardTableViewCell.m
```diff
@@ -45,7 +45,12 @@ -(void)setTestSuite:(AbstractSuite*)testSuite{
     
     [self.titleLabel setText:[LocalizationUtility getNameForTest:testSuite.name]];
     [self.descLabel setText:[LocalizationUtility getDescriptionForTest:testSuite.name]];
-    NSString *time = NSLocalizedFormatString(@"Dashboard.Card.Seconds", [NSString stringWithFormat:@"%d", [testSuite getRuntime]]);
+    //TODO convert seconds to minutes and hours when needed
+    //if getRuntime = 0 show one hour
+    int runTime = [testSuite getRuntime];
+    if (runTime == [MAX_RUNTIME_DISABLED intValue])
+        runTime = SECONDS_IN_HOUR;
+    NSString *time = NSLocalizedFormatString(@"Dashboard.Card.Seconds", [NSString stringWithFormat:@"%d", runTime]);
```
bassosimone: It seems this code is duplicate, so maybe we can factor it?

ooniprobe/Test/Test/WebConnectivity.mm
```diff
@@ -18,6 +18,7 @@ -(void) runTest {
         [TestUtility downloadUrls:^(NSArray *urls) {
             [self setUrls:urls];
             [self setDefaultMaxRuntime];
+            [[NSNotificationCenter defaultCenter] postNotificationName:@"updateRuntime" object:nil];
```
bassosimone: Do I understand correctly that here you're sending a message to the UI thread to draw the runtime?

ooniprobe/Test/Test/WebConnectivity.mm
lorenzoPrimi: I prefer to use a constant (-1) to indicate the disabled. The user can't set lower than 10.

ooniprobe/Test/Test/WebConnectivity.mm
lorenzoPrimi: I will document it

ooniprobe/View/RunTest/TestOverviewViewController.m
lorenzoPrimi: There is  no way you can set a number lower than 10  in  the  max_runtime var. 
Or you can set it to MAX_RUNTIME_DISABLED (-1) with a switch

ooniprobe/View/RunTest/TestRunningViewController.mm
lorenzoPrimi: same above

ooniprobe/Test/Test/WebConnectivity.mm
lorenzoPrimi: I tell the UI that something has happened (finished downloading url list) so the runtime should be updated. In this case written more than updated

ooniprobe/View/TestResults/Rows/DashboardTableViewCell.m
lorenzoPrimi: ok

ooniprobe/Test/Test/WebConnectivity.mm
```diff
@@ -61,8 +69,12 @@ -(void)setUrls:(NSArray*)inputs{
...
 
 -(void)setDefaultMaxRuntime {
-    NSNumber *max_runtime = [[NSUserDefaults standardUserDefaults] objectForKey:@"max_runtime"];
-    self.settings.options.max_runtime = max_runtime;
+    if ([[[NSUserDefaults standardUserDefaults] objectForKey:@"max_runtime_enabled"] boolValue]){
+        NSNumber *max_runtime = [[NSUserDefaults standardUserDefaults] objectForKey:@"max_runtime"];
+        self.settings.options.max_runtime = max_runtime;
+    }
+    else
+        self.settings.options.max_runtime = MAX_RUNTIME_DISABLED;
```
bassosimone: So, reading ObjC code is easier. Why are we caching the `max_runtime` from settings to an attribute of the WebConnectivity class? What is the purpose?

lorenzoPrimi: because a test has a setting class assigned

PR #327, opened 2020-01-13
Remove logs weekly

lorenzoPrimi:
Fixes https://github.com/ooni/probe/issues/928

## Proposed Changes

  - When we call delete log function we check if a week passed between the test date and the actual date


bassosimone:
@lorenzoPrimi Travis is currently broken because the code does not compile. Can you take a look?

https://travis-ci.org/ooni/probe-ios/builds/636196966#L1164

PR #328, opened 2020-01-15
Load text smoothly

lorenzoPrimi:
Fixes https://github.com/ooni/probe-android/issues/229 and https://github.com/ooni/probe/issues/865

## Proposed Changes

  - Saving the text in a strong variable `self.text` so we don't have to read the TextView when copying to clipboard
  - Write on the TextView on UIThread
  - Initialise the `TextView` with a string saying "Loading..."


lorenzoPrimi:
It's explained in the PR description, point N2.
Avg. loading time 5-6 seconds with a "Loading" string preloaded into the textview

PR #335, opened 2020-01-30
New algorithm to calculate runtime

lorenzoPrimi:
Fixes https://github.com/ooni/probe/issues/931

## Proposed Changes

  - Dropping runtime from `Result` object in  the database
  - Calculating runtime at runtime with this formula `result_runtime = last_row_start_time - first_row_start_time + last_row_runtime`

### DIFF
--- a/ooniprobe/Model/Database/Result.mm
+++ b/ooniprobe/Model/Database/Result.mm
@@ -1,6 +1,5 @@
 #import "Result.h"
 #import "TestUtility.h"
-#import <Crashlytics/Crashlytics.h>
 
 @implementation Result
 @dynamic test_group_name, start_time, network_id, is_viewed, is_done, data_usage_up, data_usage_down;
@@ -110,8 +109,6 @@
     Measurement *first = [self getFirstMeasurement];
     Measurement *last = [self getLastMeasurement];
     if (first == nil || last == nil){
-        //TODO move this code to a ExceptionRecord class. #983
-        [CrashlyticsKit recordError:[NSError errorWithDomain:@"get_runtime_null" code:0 userInfo:nil]];
         return ERROR_RUNTIME;
     }
     NSTimeInterval secondsBetweenTests = [last.start_time timeIntervalSinceDate:first.start_time];
--- a/ooniprobe/Model/Database/Result.mm
+++ b/ooniprobe/Model/Database/Result.mm
@@ -1,5 +1,6 @@
 #import "Result.h"
 #import "TestUtility.h"
+#import <Crashlytics/Crashlytics.h>
 
 @implementation Result
 @dynamic test_group_name, start_time, network_id, is_viewed, is_done, data_usage_up, data_usage_down;
@@ -108,8 +109,11 @@
 -(float)getRuntime{
     Measurement *first = [self getFirstMeasurement];
     Measurement *last = [self getLastMeasurement];
-    if (first == nil || last == nil)
+    if (first == nil || last == nil){
+        //TODO move this code to a ExceptionRecord class. #983
+        [CrashlyticsKit recordError:[NSError errorWithDomain:@"get_runtime_null" code:0 userInfo:nil]];
         return 0;
+    }
     NSTimeInterval secondsBetweenTests = [last.start_time timeIntervalSinceDate:first.start_time];
     return secondsBetweenTests + last.runtime;
 }
--- a/ooniprobe/Model/Database/Result.mm
+++ b/ooniprobe/Model/Database/Result.mm
@@ -108,6 +108,8 @@
 -(float)getRuntime{
     Measurement *first = [self getFirstMeasurement];
     Measurement *last = [self getLastMeasurement];
+    if (first == nil || last == nil)
+        return 0;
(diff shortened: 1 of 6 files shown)


ooniprobe/Model/Database/Result.mm
```diff
@@ -90,8 +83,25 @@ -(NSString*)getLocalizedNetworkType{
...
+
+-(Measurement*)getLastMeasurement{
+    SRKResultSet *measurements = [[[[Measurement query] where:@"result_id = ? AND is_rerun = 0" parameters:@[self]] orderByDescending:@"start_time"] fetch];
+    if ([measurements count] > 0)
+        return [measurements objectAtIndex:0];
+    return nil;
+}
+
+-(float)getRuntime{
+    Measurement *first = [self getFirstMeasurement];
```
hellais: What happens if you get `nil` here? Will it not trigger a NULL pointer exception? Can we maybe handle that case a bit more gracefully and perhaps set a fixed maxRuntime if that happens?

ooniprobe/Model/Database/Result.mm
```diff
@@ -90,8 +83,25 @@ -(NSString*)getLocalizedNetworkType{
...
-    self.runtime+=value;
+-(Measurement*)getFirstMeasurement{
+    SRKResultSet *measurements = [[[[Measurement query] where:@"result_id = ? AND is_rerun = 0" parameters:@[self]] order:@"start_time"] fetch];
+    if ([measurements count] > 0)
+        return [measurements objectAtIndex:0];
+    return nil;
+}
+
+-(Measurement*)getLastMeasurement{
+    SRKResultSet *measurements = [[[[Measurement query] where:@"result_id = ? AND is_rerun = 0" parameters:@[self]] orderByDescending:@"start_time"] fetch];
```
hellais: Can you set a LIMIT on the query so that you don't retrieve all the rows just to get the first or last one?

ooniprobe/Model/Database/Result.mm
```diff
@@ -90,8 +83,25 @@ -(NSString*)getLocalizedNetworkType{
...
+    if ([measurements count] > 0)
+        return [measurements objectAtIndex:0];
+    return nil;
+}
+
+-(float)getRuntime{
+    Measurement *first = [self getFirstMeasurement];
+    Measurement *last = [self getLastMeasurement];
+    NSTimeInterval secondsBetweenTests = [last.start_time timeIntervalSinceDate:first.start_time];
+    return secondsBetweenTests + first.runtime;
```
hellais: You should be summing the runtime of the **last** run to get the total runtime.

ooniprobe/Model/Database/Result.mm
lorenzoPrimi: You can't get nil because you can't access the test summary screen if there are no measurement in the result. See https://github.com/ooni/probe/issues/855

ooniprobe/Model/Database/Result.mm
lorenzoPrimi: Great, will do

ooniprobe/Model/Database/Result.mm
lorenzoPrimi: Correct

ooniprobe/Model/Database/Result.mm
hellais: The fact you cannot have this be `nil` because of where you call the `getRuntime` in the code is something that may change in the future and you are guaranteed to forget to add the safety then.

Please practice good defensive coding and add checks for `nil`. I think it's fine to also just handle it as an exception, but it should be explicited.

lorenzoPrimi: Would a 
```
    if (first == nil || last == nil)
        return 0;
```
work?

hellais: What pattern is being used elsewhere for handling null pointer exceptions?

lorenzoPrimi: We are not using exceptions at all in iOS. They are not commonly used.

ooniprobe/Model/Database/Result.mm
```diff
@@ -108,6 +108,8 @@ -(Measurement*)getLastMeasurement{
 -(float)getRuntime{
     Measurement *first = [self getFirstMeasurement];
     Measurement *last = [self getLastMeasurement];
+    if (first == nil || last == nil)
+        return 0;
```
hellais: I don't think it's a good idea to return a valid runtime value. I think it should rather be treated as an error and logged as such.
My suggestion is to throw an exception which mentions explicitly what is going on, see: https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/Exceptions/Tasks/HandlingExceptions.html.

lorenzoPrimi: Even if I throw an exception I still need to return a value. We can't make the app crash

hellais: I think that we need to record this somehow as an unexpected event that needs to be investigated as it's probably hiding some bugs in the app. Making the app crash (which is what would happen in the case of a NULL pointer exception) with a specific exception is better than hiding a bug.

PR #337, opened 2020-02-05
Custom websites: sanitize input

lorenzoPrimi:
Fixes https://github.com/ooni/probe/issues/982

### DIFF
--- a/ooniprobe/View/Settings/CustomURLViewController.m
+++ b/ooniprobe/View/Settings/CustomURLViewController.m
@@ -44,8 +44,9 @@
 -(IBAction)run:(id)sender{
     urlArray = [[NSMutableArray alloc] init];
     for (NSString *url in self.urlsList){
-        if (![url isEqualToString:@"http://"] && ![url isEqualToString:@"https://"]){
-            Url *currentUrl = [Url checkExistingUrl:url];
+        NSString *sanitizedUrl= [url stringByTrimmingCharactersInSet:[NSCharacterSet newlineCharacterSet]];
+        if (![sanitizedUrl isEqualToString:@"http://"] && ![sanitizedUrl isEqualToString:@"https://"]){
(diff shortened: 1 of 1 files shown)


ooniprobe/View/Settings/CustomURLViewController.m
```diff
@@ -44,8 +44,9 @@ -(IBAction)addRow:(id)sender{
 -(IBAction)run:(id)sender{
     urlArray = [[NSMutableArray alloc] init];
     for (NSString *url in self.urlsList){
-        if (![url isEqualToString:@"http://"] && ![url isEqualToString:@"https://"]){
-            Url *currentUrl = [Url checkExistingUrl:url];
+        NSString *sanitizedUrl= [url stringByTrimmingCharactersInSet:[NSCharacterSet newlineCharacterSet]];
+        if (![sanitizedUrl isEqualToString:@"http://"] && ![sanitizedUrl isEqualToString:@"https://"]){
```
bassosimone: This is slightly different from the Android algorithm. Maybe we should use the same algorithm in both cases? What is the reason why here you've introduced this extra check?

lorenzoPrimi: On iOS we have this nice `stringByTrimmingCharactersInSet` that takes a `NSCharacterSet` but on Android we are forced to use a regExp (that I don't personally like)

bassosimone: The question is why here you check for `http://` and `https://` and on Android you don't

lorenzoPrimi: On Android we use `Patterns.WEB_URL.matcher(value).matches()` that I believe does the same thing

bassosimone: you're right, thanks!

PR #339, opened 2020-02-18
Created exception utility

lorenzoPrimi:
Fixes https://github.com/ooni/probe/issues/983 

## Proposed Changes
  - Created a class called ExceptionUtility that imports Crashlytics and sends crash to them.
In this way if we decide to remove Crashlytics and use any other crash reporting system in the code, it will be a one line edit.

### DIFF
--- a/ooniprobe/Utility/ExceptionUtility.h
+++ b/ooniprobe/Utility/ExceptionUtility.h
@@ -1,0 +1,10 @@
+#import <Foundation/Foundation.h>
+#import <Crashlytics/Crashlytics.h>
+
+NS_ASSUME_NONNULL_BEGIN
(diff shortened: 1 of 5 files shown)


ooniprobe/Utility/ExceptionUtility.h
```diff
@@ -0,0 +1,12 @@
+#import <Foundation/Foundation.h>
+#import <Crashlytics/Crashlytics.h>
```
bassosimone: I believe you should import this header inside the implementation, not inside the interface header

lorenzoPrimi: done

PR #343, opened 2020-03-05
Interrupt test

lorenzoPrimi:
Fixes https://github.com/ooni/probe/issues/1026

## Proposed Changes
- Adding an X button in the running screen
- If the test is interruptable send the signal to the go library
- Empty next tests queue


ooniprobe/Test/Suite/AbstractSuite.h
```diff
@@ -13,6 +13,7 @@
 @property NSMutableArray *testList;
 @property int measurementIdx;
 @property (nonatomic) UIBackgroundTaskIdentifier backgroundTask;
+@property BOOL interrupted;
```
bassosimone, line 16: I assume the default is `atomic`? Maybe, can you make it explicitly `atomic`?

ooniprobe/Test/Test/AbstractTest.h
```diff
@@ -25,6 +26,7 @@
 @property Settings *settings;
 @property BOOL annotation;
 @property dispatch_queue_t serialQueue;
+@property id<ExperimentTask> task;
```
bassosimone, line 29: Also this property should be `atomic`

ooniprobe/Test/Suite/AbstractSuite.h
lorenzoPrimi: There is no need to, unless we want `weak`

ooniprobe/Test/Test/AbstractTest.h
lorenzoPrimi: Same as above

PR #377, opened 2020-07-08
Check and delete Uploaded report files with the new api

lorenzoPrimi:
Fixes https://github.com/ooni/probe/issues/1195

The algorithm would be:
- Query DB for measurement uploaded and with report_id
- Cycle them to get the ones with json file
- UNIQUE the report_id
- Call the api for every report_id
- If true 
    - query db to get all measurement with that report_id
    - Delete json


OONIProbeUnitTests/APITest.m
```diff
@@ -76,30 +76,20 @@ -(void)testDownloadUrls{
...
-                    [expectation fulfill];
-                }
-            }
-        } else {
-            XCTAssert(false);
+    [OONIApi checkReportId:EXISTING_REPORT_ID onSuccess:^(BOOL found){
+        if (found){
+            XCTAssert(true);
             [expectation fulfill];
+
```
bassosimone, line 86: Please rewrite as

```
XCTAssert(found);
```

we're not handling the case where found is false.

ooniprobe/Model/Database/Measurement.mm
```diff
@@ -112,61 +118,9 @@ -(void)deleteObject{
...
-        ![[resultsArray objectAtIndex:0] objectForKey:@"measurement_url"]) {
-        errorcb([NSError errorWithDomain:@"io.ooni.api"
-                                    code:ERR_JSON_EMPTY
-                                userInfo:@{NSLocalizedDescriptionKey:@"Modal.Error.JsonEmpty"
-                                           }]);
-        return;
-    }
-    successcb([[resultsArray objectAtIndex:0] objectForKey:@"measurement_url"]);
+    [OONIApi getExplorerUrl:self.report_id
+                    withUrl:self.url_id!=nil?self.url_id.url:nil
```
bassosimone, line 122: This could be simplified as: `self.url_id`

lorenzoPrimi: I don't need `self.url_id` but `self.url_id.url` that is a value inside the `Url` object

bassosimone: Ah, so the code was not readable, please: `((self.url_id != nil) ? self.url_id.url : nil)`

lorenzoPrimi: Got it! done

PR #384, opened 2020-08-21
Podfile: use ooni/probe-engine 0.16.0 for iOS

bassosimone:
We also switch to using the WebConnectivity provided by probe-engine.

Yet, we're still using MK for some tasks, so we can't pull the plug for now.

Part of https://github.com/ooni/probe-engine/issues/748.

### DIFF
--- a/Podfile
+++ b/Podfile
@@ -2,7 +2,7 @@ platform :ios, '9.0'
 # ignore all warnings from all pods
 inhibit_all_warnings!
 target 'ooniprobe' do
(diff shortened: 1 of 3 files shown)


lorenzoPrimi:
I tested it and works, do you want it to land in master or do more testing yourself?

bassosimone:
@lorenzoPrimi did you try running w/o time limits? If that worked, I guess we can land! Thank you!

lorenzoPrimi (approved):

PR #392, opened 2020-10-06
engine: remove MK dependency

lorenzoPrimi:
Part of https://github.com/ooni/probe-engine/issues/668


ooniprobe/Engine/OONISessionConfig.h
```diff
@@ -0,0 +1,58 @@
+#import <Foundation/Foundation.h>
+#import <oonimkall/Oonimkall.h>
+#import "OONILogger.h"
+
+/** OONISessionConfig contains configuration for a OONISession. */
+@interface OONISessionConfig : NSObject
```
bassosimone, line 6:
```suggestion
@interface OONISessionConfig : NSObject

```

ooniprobe/OONILogger.h
```diff
@@ -0,0 +1,13 @@
...
+//
+//  Created by Lorenzo Primiterra on 05/10/2020.
+//  Copyright © 2020 OONI. All rights reserved.
+//
+
+#ifndef OONILogger_h
+#define OONILogger_h
+
+
+#endif /* OONILogger_h */
```
bassosimone, line 13: It seems we don't need this file

ooniprobe/OONIMKTask.h
```diff
@@ -0,0 +1,13 @@
...
+//
+//  Created by Lorenzo Primiterra on 05/10/2020.
+//  Copyright © 2020 OONI. All rights reserved.
+//
+
+#ifndef OONIMKTask_h
+#define OONIMKTask_h
+
+
+#endif /* OONIMKTask_h */
```
bassosimone, line 13: It seems we don't need this file

ooniprobe/OONIMKTaskConfig.h
```diff
@@ -0,0 +1,13 @@
...
+//
+//  Created by Lorenzo Primiterra on 05/10/2020.
+//  Copyright © 2020 OONI. All rights reserved.
+//
+
+#ifndef OONIMKTaskConfig_h
+#define OONIMKTaskConfig_h
+
+
+#endif /* OONIMKTaskConfig_h */
```
bassosimone, line 13: It seems we don't need this file

ooniprobe/OONISession.h
```diff
@@ -0,0 +1,13 @@
...
+//
+//  Created by Lorenzo Primiterra on 05/10/2020.
+//  Copyright © 2020 OONI. All rights reserved.
+//
+
+#ifndef OONISession_h
+#define OONISession_h
+
+
+#endif /* OONISession_h */
```
bassosimone, line 13: It seems we don't need this file.

ooniprobe/Utility/OONIApi.m
```diff
@@ -3,16 +3,21 @@
...
-    if ([results isGood])
-        cc = [results probeCC];
+
+    cc = [Engine resolveProbeCCwithSoftwareName:SOFTWARE_NAME
+                                          softwareVersion:[VersionUtility get_software_version]
+                                                  timeout:DEFAULT_TIMEOUT
+                                                    error:&error];
+    if (error != nil) {
+        //TODO Do we care about the error?
+    }
```
bassosimone, line 20:
```suggestion
```

We don't.

ooniprobe/Engine/Engine.m
```diff
@@ -1,40 +1,90 @@
...
+                                                                   logger:[LoggerNull new]]
+                                                                    error:error];
+    // Updating resources with no timeout because we don't know for sure how much
+    // it will take to download them and choosing a timeout may prevent the operation
+    // to ever complete. (Ideally the user should be able to interrupt the process
+    // and there should be no timeout here.)
+    [session maybeUpdateResources:[session newContext]
+                            error:error];
+    return [session geolocate:[session newContextWithTimeout:timeout]
+                 error:error].country;
```
bassosimone, line 34: I think we cannot be as compact as we were in Java because here there are errors rather than exceptions:

```suggestion
/** resolveProbeCC returns the probeCC. */
+ (NSString*) resolveProbeCCwithSoftwareName:(NSString*)softwareName
                             softwareVersion:(NSString*)softwareVersion
                                     timeout:(long)timeout
                                       error:(NSError **)userError {
    NSError *___ = nil;
    NSError **error = (userError != nil) ? userError : &___;
    PESession* session = [[PESession alloc] initWithConfig:
                          [Engine getDefaultSessionConfigWithSoftwareName:softwareName
                                                          softwareVersion:softwareVersion
                                                                   logger:[LoggerNull new]]
                                                                    error:error];
    if (error != nil) {
        return nil;
    }
    // Updating resources with no timeout because we don't know for sure how much
    // it will take to download them and choosing a timeout may prevent the operation
    // to ever complete. (Ideally the user should be able to interrupt the process
    // and there should be no timeout here.)
    [session maybeUpdateResources:[session newContext] error:error];
    if (error != nil) {
        return nil;
    }
    OONIGeolocateResults *r = [session geolocate:[session newContextWithTimeout:timeout] error:error];
    if (error != nil) {
        return nil;
    }
    return r.country;
```

ooniprobe/Engine/PESession.m
```diff
@@ -0,0 +1,39 @@
...
+    if (self) {
+        self.session = OonimkallNewSession([config toOonimkallSessionConfig], error);
+    }
+    return self;
+}
+
+- (OONIGeolocateResults*) geolocate:(OONIContext*) ctx error:(NSError **)error{
+    return [[OONIGeolocateResults alloc] initWithResults:
+            [self.session geolocate:ctx.ctx error:error]];
+
```
bassosimone, line 16:
```suggestion
    OonimkallGeolocateResults *r =  [self.session geolocate:ctx.ctx error:error];
    return (r != nil) ? ([[OONIGeolocateResults alloc] initWithResults:r) : nil;
```

ooniprobe/Engine/PESession.m
```diff
@@ -0,0 +1,39 @@
...
+}
+
+- (void) maybeUpdateResources:(OONIContext*) ctx error:(NSError **)error{
+    [self.session maybeUpdateResources:ctx.ctx
+                                 error:error];
```
bassosimone, line 21:
```suggestion
    [self.session maybeUpdateResources:ctx.ctx error:error];
```

ooniprobe/Engine/PESession.m
```diff
@@ -0,0 +1,39 @@
...
+}
+
+- (OONIContext*) newContextWithTimeout:(long)timeout{
+    return [[OONIContext alloc] initWithContext:
+            [self.session newContextWithTimeout:timeout]];
+
+}
+
+- (OONISubmitResults*) submit:(OONIContext*) ctx measurement:(NSString*) measurement error:(NSError **)error{
+    return [[OONISubmitResults alloc] initWithResults:[self.session submit:ctx.ctx measurement:measurement error:error]];
```
bassosimone, line 35:
```suggestion
    OonimkallSubmitResults *r = self.session submit:ctx.ctx measurement:measurement error:error];
    return (r != nil) ? ([[OONISubmitResults alloc] initWithResults:r]) : nil;
```

ooniprobe/View/TestResults/Footer/UploadFooterViewController.m
```diff
@@ -79,22 +81,38 @@ -(void)uploadMeasurements:(NSArray *)notUploaded{
...
+                                                                        error:&error];
+        if (error != nil) {
+            //TODO Do we care about the error?
+        }
+        // Updating resources with no timeout because we don't know for sure how much
+        // it will take to download them and choosing a timeout may prevent the operation
+        // to ever complete. (Ideally the user should be able to interrupt the process
+        // and there should be no timeout here.)
+        [session maybeUpdateResources:[session newContext]
+                                error:&error];
```
bassosimone, line 103:
```suggestion
        [session maybeUpdateResources:[session newContext] error:&error];
```

ooniprobe/View/TestResults/Footer/UploadFooterViewController.m
```diff
@@ -79,22 +81,38 @@ -(void)uploadMeasurements:(NSArray *)notUploaded{
...
-        id<CollectorTask> task = [Engine collectorTaskWithSoftwareName:SOFTWARE_NAME
-                               softwareVersion:[VersionUtility get_software_version]];
+        NSError *error;
+        PESession* session = [[PESession alloc] initWithConfig:
+                              [Engine getDefaultSessionConfigWithSoftwareName:SOFTWARE_NAME
+                                                              softwareVersion:[VersionUtility get_software_version]
+                                                                       logger:self.logger]
+                                                                        error:&error];
+        if (error != nil) {
+            //TODO Do we care about the error?
```
bassosimone, line 96:
```suggestion
            //TODO Do we care about the error?
            return;
```

ooniprobe/View/TestResults/Footer/UploadFooterViewController.m
```diff
@@ -79,22 +81,38 @@ -(void)uploadMeasurements:(NSArray *)notUploaded{
...
+        }
+        // Updating resources with no timeout because we don't know for sure how much
+        // it will take to download them and choosing a timeout may prevent the operation
+        // to ever complete. (Ideally the user should be able to interrupt the process
+        // and there should be no timeout here.)
+        [session maybeUpdateResources:[session newContext]
+                                error:&error];
+        //TODO what to do with error?
+        if (error != nil) {
+            //TODO Do we care about the error?
```
bassosimone, line 106:
```suggestion
            //TODO Do we care about the error?
            return;
```

bassosimone (changes requested):
I have added as many comments as possible. Whenever it made sense, I also provided a concrete suggestion highlighting how I suggest to change the code. I may still need another pass once these changes are applied (or rejected w/ motivation) to double check error handling.

Thank you for keeping this large diff tidy and understandable 🎸

ooniprobe/Engine/PESession.m
lorenzoPrimi: Why this?

ooniprobe/View/TestResults/Footer/UploadFooterViewController.m
lorenzoPrimi: return in the first if or second?

bassosimone: I think we should return in any case there is an error.

lorenzoPrimi: even for `maybeUpdateResources` ? is it a problem running a test with old resources?

ooniprobe/Engine/PESession.m
lorenzoPrimi: We never check if the returned `OonimkallSubmitResults` but we only check the error

ooniprobe/View/TestResults/Footer/UploadFooterViewController.m
bassosimone: Yes, I think we should return. There are at least reasons for this:

1. the accuracy of measurements depends on the resources being up to date (and we have had data quality incidents in the past caused by assets not being up-to-date;

2. if you eventually ignore the error, the code in probe-engine that will execute later will just try again to fetch the resources and fail, so there is no easy escape from fetching the resources;

3. while it could be tempting to rely on the fact that probe-engine would try to fetch the resources, I'd rather not do that, because that would lead to context mismatches (e.g. we have a 30 seconds deadline for finding the location and, if we use such deadline for possibly downloading the resources and finding the location, certain clients may loop because they cannot really download the resources in 30 seconds).

ooniprobe/Engine/PESession.m
bassosimone: While that is true, the typical Apple convention is to return either `nil` with an error or a valid value. Because that convention is used by many system APIs as well as APIs maintained by us, not using this convention here is surprising and fragile. For example, if I was to change the code, I would most likely assume that the API only returns `nil` in case of error. This is even more important if we consider that the error argument might in principle be `nil`, leading a developer with no way of assessing whether the result is successful.

ooniprobe/Engine/PESession.m
```diff
@@ -0,0 +1,38 @@
...
+- (id)initWithConfig:(OONISessionConfig *)config error:(NSError **)error{
+    self = [super init];
+    if (self) {
+        self.session = OonimkallNewSession([config toOonimkallSessionConfig], error);
+    }
+    return self;
+}
+
+- (OONIGeolocateResults*) geolocate:(OONIContext*) ctx error:(NSError **)error{
+    OonimkallGeolocateResults *r =  [self.session geolocate:ctx.ctx error:error];
```
bassosimone, line 14:
```suggestion
    OonimkallGeolocateResults *r = [self.session geolocate:ctx.ctx error:error];
```

ooniprobe/Engine/Engine.h
```diff
@@ -1,26 +1,42 @@
...
 
-+ (id<ExperimentTask>) startExperimentTaskWithSettings:(id<ExperimentSettings>)settings error:(NSError **)error;
++ (id<OONIMKTask>) startExperimentTaskWithSettings:(id<OONIMKTaskConfig>)settings error:(NSError **)error;
+
++ (NSString*) resolveProbeCCWithSoftwareName:(NSString*)softwareName
+                             softwareVersion:(NSString*)softwareVersion
+                                     timeout:(long)timeout
+                                       error:(NSError **)error;
+
++ (PESession*) newSession:(OONISessionConfig*) config error:(NSError **)error;
```
bassosimone, line 30:
```suggestion
+ (id<OONISession>) newSession:(OONISessionConfig*) config error:(NSError **)error;
```

We probably want to avoid exposing `PESession`?

ooniprobe/Engine/Engine.m
```diff
@@ -1,40 +1,90 @@
...
+                                                                    error:error];
+    // Updating resources with no timeout because we don't know for sure how much
+    // it will take to download them and choosing a timeout may prevent the operation
+    // to ever complete. (Ideally the user should be able to interrupt the process
+    // and there should be no timeout here.)
+    [session maybeUpdateResources:[session newContext]
+                            error:error];
+    if (error != nil)
+        return nil;
+    return [session geolocate:[session newContextWithTimeout:timeout] error:error].country;
```
bassosimone, line 34: Error handling here seems incorrect: `error` is non `nil` unless the user passes a `nil` `error` argument. I think we should restructure this function as follows to address all possible cases:

```suggestion
/** resolveProbeCC returns the probeCC. */
+ (NSString*) resolveProbeCCWithSoftwareName:(NSString*)softwareName
                             softwareVersion:(NSString*)softwareVersion
                                     timeout:(long)timeout
                                       error:(NSError **)userError {
    NSError *defaultError;
    NSError **error = (userError != nil) ? userError : &defaultError;
    PESession* session = [[PESession alloc] initWithConfig:
                          [Engine getDefaultSessionConfigWithSoftwareName:softwareName
                                                          softwareVersion:softwareVersion
                                                                   logger:[LoggerNull new]]
                                                                    error:error];
    if (*error != nil) {
        return nil;
    }
    // Updating resources with no timeout because we don't know for sure how much
    // it will take to download them and choosing a timeout may prevent the operation
    // to ever complete. (Ideally the user should be able to interrupt the process
    // and there should be no timeout here.)
    [session maybeUpdateResources:[session newContext]
                            error:error];
    if (*error != nil) {
        return nil;
    }
    return [session geolocate:[session newContextWithTimeout:timeout] error:error].country;
```

ooniprobe/Utility/OONIApi.m
```diff
@@ -3,16 +3,19 @@
...
-    id<GeoIPLookupResults> results = [task perform];
-    NSString *cc = @"XX";
-    if ([results isGood])
-        cc = [results probeCC];
+    NSError *error;
+    NSString *cc = [Engine resolveProbeCCWithSoftwareName:SOFTWARE_NAME
+                                          softwareVersion:[VersionUtility get_software_version]
+                                                  timeout:DEFAULT_TIMEOUT
+                                                    error:&error];
+    if (error != nil || cc == nil) {
```
bassosimone, line 16: In this context it's sufficient to check for one of them, not both. I'd vote for error:

```suggestion
    if (error != nil) {
```

ooniprobe/View/TestResults/Footer/UploadFooterViewController.m
```diff
@@ -79,22 +81,37 @@ -(void)uploadMeasurements:(NSArray *)notUploaded{
...
+                                                                        error:&error];
+        if (error != nil) {
+            return;
+        }
+        // Updating resources with no timeout because we don't know for sure how much
+        // it will take to download them and choosing a timeout may prevent the operation
+        // to ever complete. (Ideally the user should be able to interrupt the process
+        // and there should be no timeout here.)
+        [session maybeUpdateResources:[session newContext] error:&error];
+        //TODO what to do with error?
```
bassosimone, line 103:
```suggestion
        //TODO what to do with error?
```

I think the error is now handled, isn't it?

ooniprobe/Engine/Engine.m
```diff
@@ -1,40 +1,90 @@
...
-    }
-    return [[MKExperimentTaskAdapter alloc]
-            initWithTask:[MKAsyncTask start:settings.dictionary]];
+/** startExperimentTaskWithSettings starts the experiment described by the provided settings. */
++ (id<OONIMKTask>) startExperimentTaskWithSettings:(id<OONIMKTaskConfig>)settings error:(NSError **)error{
+    return [[PEMKTask alloc] initWithTask:OonimkallStartTask(settings.serialization, error)];
+}
+
+/** resolveProbeCC returns the probeCC. */
++ (NSString*) resolveProbeCCWithSoftwareName:(NSString*)softwareName
```
bassosimone, line 17: I am not fully convinced of how errors propagate throughout this function and whether errors may as well be overwritten. I proposed some code that seems to handle the case of errors.

bassosimone (changes requested):
I've marked the only place in which I am still not convinced 🥋

ooniprobe/Utility/OONIApi.m
lorenzoPrimi: are we sure about it?
if cc is null it would crash big times the app, I prefer the double check as cc doesn't have a default value

bassosimone: So, I'd rather check for just `cc` then! (All of this is complex just because we don't have union types 😢)

bassosimone (approved):
🐳

PR #406, opened 2020-11-23
Fixing database query that prevented json deletion

lorenzoPrimi:
Discovered in https://github.com/ooni/probe/issues/1271

### DIFF
--- a/ooniprobe/Model/Database/Measurement.m
+++ b/ooniprobe/Model/Database/Measurement.m
@@ -18,8 +18,10 @@
 }
 
 + (SRKResultSet*)selectWithReportId:(NSString*)report_id {
-    return [[[Measurement query] where:
-             [NSString stringWithFormat:@"%@ AND report_id = '%@'", REPORT_QUERY, report_id]] fetch];
+    return [[[Measurement query]
+               where:[NSString stringWithFormat:@"%@ AND report_id = ?", REPORT_QUERY]
+               parameters:@[report_id]]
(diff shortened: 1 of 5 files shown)


ooniprobe/Model/Database/Measurement.m
```diff
@@ -19,12 +19,12 @@ + (SRKResultSet*)notUploadedMeasurements {
 
 + (SRKResultSet*)selectWithReportId:(NSString*)report_id {
     return [[[Measurement query] where:
-             [NSString stringWithFormat:@"%@ AND report_id = %@", UPLOADED_QUERY, report_id]] fetch];
+             [NSString stringWithFormat:@"%@ AND report_id = '%@'", REPORT_QUERY, report_id]] fetch];
```
bassosimone, line 22: Is there a library we can use to construct queries instead of concatenating strings?

bassosimone (changes requested):
I think this issue here should make us reflect that we need better tools to construct queries.

🤔

ooniprobe/Model/Database/Measurement.m
lorenzoPrimi: Where takes a string, so you need to construct the string somehow. I prefer to use *define* strings like `UPLOADED_QUERY` so I don't have to copy it around

bassosimone (approved):
🐳 

(My previous request for changes was actually wrong!)

PR #408, opened 2020-12-04
cleanup(oonimkall): remove legacy option

lorenzoPrimi:
Fixes https://github.com/ooni/probe/issues/1305
```
report_id\“:\“20201207T163355Z_facebookmessenger_PT_12353_n1_C0IKIV6CxggHUd8E\“,
```


bassosimone:
@lorenzoPrimi it seems tests are failing quite badly. This does not seems to be just a problem with GitHub Actions, but rather it seems to appear also when I run tests locally. One symptom for this kind of failure is that tests run for ~5 hours. I have been running them locally for several minutes now and they spinning on:

```
2020-12-07 13:51:30.231650+0100 ooniprobe[27048:4644512] DB error: attempt to write a readonly database
2020-12-07 13:51:30.231997+0100 ooniprobe[27048:4644512] DB error: attempt to write a readonly database
2020-12-07 13:51:30.232605+0100 ooniprobe[27048:4644512] DB error: attempt to write a readonly database
```

My computer is heating a lot, so it really seems the app is spinning. Judging by when the tests specifically started to fail, it seems there are two good candidate commits:

1. https://github.com/ooni/probe-ios/commit/5806827af334c3fccbc6cce0490081b6cc9d3842

2. https://github.com/ooni/probe-ios/commit/cf80b2862675c4cc0c8be4d457e4315fca30149c

I believe we should fix this issue before proceeding. Not being able to run tests (locally or remotely) and specifically the fact that it seems we're spinning leads me to think we should investigate what has appened.

PR #412, opened 2020-12-09
Better interrupt tests

lorenzoPrimi:
Removing the direct interrupt call on the `task` object and use notification handler that gets destroyed after every test.


lorenzoPrimi:
Tested and working

PR #413, opened 2020-12-10
Interrupt test if network changed

lorenzoPrimi:
Fixes https://github.com/ooni/probe/issues/1296

## Proposed Changes

  - Interrupt test when connectivity changes

### DIFF
--- a/ooniprobe/AppDelegate.m
+++ b/ooniprobe/AppDelegate.m
@@ -6,6 +6,7 @@
 #import "TestRunningViewController.h"
 #import "SettingsUtility.h"
 #import "CountlyUtility.h"
+#import "ReachabilityManager.h"
 
 @interface AppDelegate ()
 
@@ -28,6 +29,7 @@
     [[UIBarButtonItem appearance] setTitleTextAttributes:[NSDictionary dictionaryWithObjectsAndKeys: [UIFont fontWithName:@"FiraSans-Regular" size:16],NSFontAttributeName, nil] forState:UIControlStateNormal];
     [NavigationBarUtility setDefaults];
     [CountlyUtility initCountly];
+    [ReachabilityManager sharedManager];
     application.statusBarStyle = UIStatusBarStyleLightContent;
     [SettingsUtility incrementAppOpenCount];
     return YES;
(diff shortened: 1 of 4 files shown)


ooniprobe/AppDelegate.m
```diff
@@ -28,6 +29,7 @@ - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(
     [[UIBarButtonItem appearance] setTitleTextAttributes:[NSDictionary dictionaryWithObjectsAndKeys: [UIFont fontWithName:@"FiraSans-Regular" size:16],NSFontAttributeName, nil] forState:UIControlStateNormal];
     [NavigationBarUtility setDefaults];
     [CountlyUtility initCountly];
+    [ReachabilityManager sharedManager];
```
bassosimone, line 32: Is the purpose of this statement to create a singleton instance? If so, could you please document it?

ooniprobe/View/RunTest/TestRunningViewController.m
```diff
@@ -169,4 +168,8 @@ -(IBAction)cancelTest:(id)sender{
                             inView:self];
 }
 
+- (void)interruptTestUI {
+    [self.runningTestsLabel setText:NSLocalizedString(@"Dashboard.Running.Stopping.Title", nil)];
+    [self.logLabel setText:NSLocalizedString(@"Dashboard.Running.Stopping.Notice", nil)];
```
bassosimone, line 173: shouldn't you remove the observer here?

bassosimone (changes requested):
Thank you! I think we need a documenting comment and, also, there is a place where I am in doubt of whether we need to delete an observer. 🧇

ooniprobe/AppDelegate.m
lorenzoPrimi: Yes, to create the singleton. Just the comment will be enough?

ooniprobe/View/RunTest/TestRunningViewController.m
lorenzoPrimi: In theory I should remove the observer when the View  is dismissed. Good catch

lorenzoPrimi: Bot on the other hand, this is automatically done by the OS when it's destroyed

ooniprobe/AppDelegate.m
bassosimone: Absolutely, thank you!

ooniprobe/View/RunTest/TestRunningViewController.m
bassosimone: Okay, thanks for clarifying!

bassosimone (approved):
🐳

PR #453, opened 2021-11-23
chore: upgrade to oonimkall v3.11.0-beta.2

bassosimone:
This (experimental?) diff points to the first version of oonimkall
automatically built using the cloud.

Unfortunately, my local system is quite in a bit of a broken state at
the moment, after the macOS 12.0 system update.

I'm going to use the CI to know whether the automatically built
oonimkall works with the current probe-ios.

Reference issue https://github.com/ooni/probe/issues/1879.

### DIFF
--- a/Podfile
+++ b/Podfile
@@ -3,7 +3,7 @@ use_frameworks!
 inhibit_all_warnings!
 
 target 'ooniprobe' do
-    pod "oonimkall", :podspec => "https://github.com/ooni/probe-cli/releases/download/v3.10.0-beta.3/oonimkall.podspec"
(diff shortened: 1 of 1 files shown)


bassosimone:
@lorenzoPrimi would you mind testing this new engine release? I suspect the most important thing to test is whether we're still able to use the bundled version of Psiphon. But, generally, I think we should test everything since quite some time has passed and this is a new release. Thank you!

bassosimone:
@lorenzoPrimi please, take a look!

lorenzoPrimi:
Tested and working as expected

PR #458, opened 2021-12-14
fix: only keep arm64 and x86_64 architectures

bassosimone:
See https://github.com/ooni/probe/issues/1918 for a description
of the problem that we want to solve.

See https://github.com/ooni/probe/issues/1918#issuecomment-993383545
instead for the explanation of what's happening.

Closes https://github.com/ooni/probe/issues/1918.

While there, add a new workflow that ensures we're not breaking
the release process in the future without noticing it.

### DIFF
--- a/.github/workflows/archive.yml
+++ b/.github/workflows/archive.yml
@@ -1,0 +1,20 @@
+name: archive
+on:
+  pull_request:
+  push:
+  schedule:
+    - cron: "0 2 * * */2"
+jobs:
+  archive:
+    runs-on: macos-latest
+    steps:
+      - uses: actions/checkout@v2
+      - run: bundle install
+      - run: pod repo update
+      - run: pod install
+      - run: |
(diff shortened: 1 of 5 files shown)


bassosimone:
Tests are failing. We apparently need to do a bit more work to make them green again. (It may also be that the issue is that we're using an old version of Xcode inside the relevant Github Action.)

lorenzoPrimi:
For reference
- arm64 is the current 64-bit ARM CPU architecture, as used since the iPhone 5S and later (6, 6S, SE and 7), the iPad Air, Air 2 and Pro, with the A7 and later chips.
- armv7s (a.k.a. Swift, not to be confused with the language of the same name), being used in Apple's A6 and A6X chips on iPhone 5, iPhone 5C and iPad 4.
- armv7, an older variation of the 32-bit ARM CPU, as used in the A5 and earlier.

PR #462, opened 2022-01-31
fix: disable riseupvpn experiment

bassosimone:
Closes https://github.com/ooni/probe/issues/2000

Closes https://github.com/ooni/probe/issues/2001


aanorbel:
LGTM ✅

PR #472, opened 2022-03-09
User initiated tests should start off minimised

aanorbel:
Fixes https://github.com/ooni/probe/issues/1976

## Proposed Changes

  - Update from implementation that uses `performSegueWithIdentifier` or  `prepareForSegue` to running tests directly and reloading constraints.

### DIFF
--- a/ooniprobe/Test/RunningTest.h
+++ b/ooniprobe/Test/RunningTest.h
@@ -11,7 +11,7 @@ NS_ASSUME_NONNULL_BEGIN
 - (void)clearSuites;
 - (void)runTest;
 - (void)networkTestEnded;
-- (void)setAndRun:(NSMutableArray*)testSuites;
+- (void)setAndRun:(NSMutableArray*)testSuites inView:(UIViewController *)view;
 @property (nonatomic, strong) NSMutableArray *testSuites;
 @property (nonatomic, strong) AbstractSuite *testSuite;
 @property (nonatomic, strong) AbstractTest *testRunning;
--- a/ooniprobe/Test/RunningTest.m
+++ b/ooniprobe/Test/RunningTest.m
@@ -1,4 +1,5 @@
 #import "RunningTest.h"
+#import "ReachabilityManager.h"
 
 @implementation RunningTest
 
@@ -26,11 +27,25 @@ static RunningTest *currentTest = nil;
     }
 }
 
--(void)setAndRun:(NSMutableArray*)testSuites {
-    @synchronized(self)
-    {
-        currentTest.testSuites = testSuites;
-        [currentTest runTest];
+- (void)setAndRun:(NSMutableArray *)testSuites inView:(UIViewController *)view {
+    if (view != nil && [[ReachabilityManager sharedManager] isVPNConnected]) {
+        [MessageUtility alertVpnWithTitle:NSLocalizedString(@"Modal.DisableVPN.Title", nil)
+                                  message:NSLocalizedString(@"Modal.DisableVPN.Message", nil)
+                                   inView:view
+                             runVPNAction:^(UIAlertAction *action) {
+                                 @synchronized (self) {
+                                     currentTest.testSuites = testSuites;
+                                     [currentTest runTest];
+                                 }
+                             }
+                         disableVPNAction:nil
+
+        ];
+    } else {
+        @synchronized (self) {
+            currentTest.testSuites = testSuites;
+            [currentTest runTest];
+        }
     }
 }
 
--- a/ooniprobe/Test/Test/AbstractTest.h
+++ b/ooniprobe/Test/Test/AbstractTest.h
@@ -36,5 +36,6 @@
 -(void)prepareRun;
 -(void)runTest;
 -(int)getRuntime;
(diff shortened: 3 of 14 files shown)


lorenzoPrimi:
makes sense
🐳

aanorbel:
Noticeable delay before showing `ProgressViewController` when running `webconnectivity` tests and tests lists.
This originates from the fact that `status.started` for web connectivity is only published after a network communication is made.

I will proceed to look into this.

bassosimone:
@aanorbel I would recommend to try with `"status.queued"`, which should happen a bit earlier. Though I am not sure this is going to make such a big difference. I will test this branch on Xcode and see if I have further insights. I think it would be useful to figure out whether there is any database pause that delays starting the test.

aanorbel:
@bassosimone  I realized that in the case of the web-connectivity, it take a couple of seconds after the test is started from the UI before the `status.queued` event is received  in the app. I think this is the cause of the delay.  Here is the appropriate log showing that.

```log
2022-03-11 11:12:17.030432+0100 ooniprobe[449:21115] engine: info: sessionresolver: https://cloudflare-dns.com/dns-query... <nil>
2022-03-11 11:12:21.041183+0100 ooniprobe[449:21111] engine: info: sessionresolver: https://cloudflare-dns.com/dns-query... <nil>
2022-03-11 11:12:23.910376+0100 ooniprobe[449:21102] engine: info: session: using probe services: {Address:https://ps2.ooni.io Type:https Front:}
2022-03-11 11:12:23.910805+0100 ooniprobe[449:21102] engine: info: iplookup: using avast
2022-03-11 11:12:24.511136+0100 ooniprobe[449:21113] engine: info: sessionresolver: https://cloudflare-dns.com/dns-query... <nil>
2022-03-11 11:12:27.408178+0100 ooniprobe[449:21102] engine: info: iplookup: using ubuntu
2022-03-11 11:12:29.665457+0100 ooniprobe[449:21111] engine: info: sessionresolver: https://dns.google/dns-query... <nil>
2022-03-11 11:12:33.355209+0100 ooniprobe[449:21103] [tcp] tcp_input [C2.1:3] flags=[R] seq=1145735649, ack=0, win=0 state=LAST_ACK rcv_nxt=1145735649, snd_una=1350682551
2022-03-11 11:12:33.357388+0100 ooniprobe[449:21103] [tcp] tcp_input [C2.1:3] flags=[R] seq=1145735649, ack=0, win=0 state=CLOSED rcv_nxt=1145735649, snd_una=1350682551
2022-03-11 11:12:37.757162+0100 ooniprobe[449:21102] Got event: {
    key = "status.queued";
    value =     {
    };
}
```

bassosimone:
Thanks @aanorbel, I've created a separate issue for that here: https://github.com/ooni/probe/issues/2054.

aanorbel:
Fixed. I merged the progress view display to master before this one.

bassosimone:
@aanorbel I still see the progress bar moving back and forth during experimental tests, which suggests that we are not tracking progress correctly. I do not think this is going to be a blocker for merging this PR, but I'd like to point out we have still not fixed this problem and wanted to touch base to ensure (a) you know about this and (b) we have an issue.

bassosimone:
@aanorbel when I tried re-running webconnectivity measurements I also got a prompt that some measurements were not uploaded. So, I clicked to upload measurements while I was re-running webconnectivity. This action of mine created a re-uploading overlay that never disappeared. It seems the whole re-upload process did not even start, or, if it started, it was just stuck and never completed:

Screenshot 2022-04-07 at 10 09 22

**Figure 1**. Re-uploading overlay appears after I click "re-upload" and never makes any progress or goes away

As a result, the webconnectivity view and the results view (see next screenshot) are now completely blocked by the re-uploading message and I cannot interact with the app in those two screens:

Screenshot 2022-04-07 at 10 14 01

**Figure 2**. We cannot get rid of this overlay until we kill the app.

I needed to kill the app to be able to access the webconnectivity view or the more general results view again.

So, I can confirm that re-running measurements _causes_ the "some not uploaded" bar to appear. It seems we need to improve our data model to factor in the fact that re-running from the result screen does not imply that some measurements are not uploaded, rather they appear as such because we are running them.

See the following figure:

Screenshot 2022-04-07 at 10 16 57

**Figure 3**. Confirmation that the prompt to re-upload appears when we re-run from the result screen.

I was initially not sure whether the "some not uploaded" stripe appeared _because_ I was re-running from the result screen or whether it actually happened that some measurements were not uploaded. So, I repeated the whole set of tests and now I can confirm that there were no measurements to re-upload until I started re-running webconnectivity.

So, I think we can conclude with confidence that the "some not uploaded" stripe appears because we're actually re-running webconnectivity measurements from the result screen.

aanorbel:
@bassosimone , yes indeed a shared codebase will help resolve similar issues in one place.

I resolved the issue mentioned above https://github.com/ooni/probe-ios/pull/472#issuecomment-1091287683 in the commits that followed. Please check it out and let me know if you still have the error.

bassosimone (approved):
🐳 🙏 💯

PR #476, opened 2022-03-17
Add minimized running test indicator missing in measurement list views.

aanorbel:
Fixes https://github.com/ooni/probe/issues/1981
| .  |  . | . | . | . |
| ----------- | ----------- | ----------- | ----------- | ----------- |
| Simulator Screen Shot - iPhone 13 Pro Max - 2022-03-17 at 13 07 25 |  Simulator Screen Shot - iPhone 13 Pro Max - 2022-03-17 at 13 07 18  |  Simulator Screen Shot - iPhone 13 Pro Max - 2022-03-17 at 13 07 05  |  Simulator Screen Shot - iPhone 13 Pro Max - 2022-03-17 at 13 03 57   |  Simulator Screen Shot - iPhone 13 Pro Max - 2022-03-17 at 13 03 42  |


bassosimone:
While testing this diff, I noticed that in-progress tests appear as "failed" while they are in progress and are later updated as WAI when the test completes. We previously did not care about this fact because tests were running in the foreground but now we should probably start caring about it. If it's easy to do that, I'd fix this in the same PR. Otherwise, I think it makes sense to create an issue. Regardless of what we do here, I'd investigate if we have the same issue for Android.

Screenshot 2022-03-18 at 10 55 57

bassosimone (approved):
This seems WAI and we agreed verbally to document and tackle any extra issue regarding the view both on Android and iOS.

🐳

PR #482, opened 2022-04-06
Improve communication of which websites are tested by OONI Probe

aanorbel:
Fixes  https://github.com/ooni/probe/issues/1929 , https://github.com/ooni/probe/issues/1935

## Proposed Changes

  - Add button to load websites (`loadFromTemplateButton`) used in `web_connectivity` test in `CustomURLViewController`.
  - Fetch and display websites in webconnectivity when `loadFromTemplateButton` is clicked, making `loadFromTemplateButton` invisible when successful.

| . | . | . |
| -- | -- | -- |
|  Simulator Screen Shot - iPhone 13 Pro Max - 2022-04-06 at 21 27 44 |Simulator Screen Shot - iPhone 13 Pro Max - 2022-04-06 at 21 27 47 |  Simulator Screen Shot - iPhone 13 Pro Max - 2022-04-06 at 22 12 23 |

### DIFF
--- a/ooniprobe/Base.lproj/Localizable.strings
+++ b/ooniprobe/Base.lproj/Localizable.strings
@@ -407,6 +407,7 @@
 "Settings.Websites.CustomURL.NoURLEntered" = "No URLs entered";
 "Settings.Websites.CustomURL.Run" = "Run";
 "Settings.Websites.CustomURL.Add" = "Add website";
+"Settings.Websites.CustomURL.LoadFromTemplate" = "Load from template";
 "Settings.Websites.TestCount" = "Number of tested websites (0 means all)";
 "Settings.InstantMessaging.TestWhatsApp" = "Test WhatsApp";
 "Settings.InstantMessaging.TestTelegram" = "Test Telegram";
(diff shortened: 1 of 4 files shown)


bassosimone:
I am testing this PR. I have to comments for now:

1. the wording "Load from template" is quite unclear to me and, also, too long to fit the screen (see screenshot) with the specific emulator that I am using (iPod touch 7th generation)

2. I did not immediately noticed that it loaded the whole test list so now I have been running tests for a long time

Both aspects seem to have to do with user experience. The _original problem_ mentioned in the issues is that users would like to know _in advance_ what URLs they are going to test in a normal run. However, the solution we have presented here is to introduce new code that do different things compared to a normal run.

So, I am not sure here we are fixing the original problem, rather it seems to me we're adding a new solution. It may be that we need to tweak this new solution to actually address the original problem.

I am going to flag this comment of mine for internal discussion. My gut feeling is that solving this problem is not as straightforward as it originally looked like. I also think that the solution we currently have may be partly good to solving the problem and that the subject of the discussion is not much the code itself rather what we want to do.

Screenshot 2022-04-07 at 11 25 25

**Figure 1** "Load from template" being too long for the emulate device.

Screenshot 2022-04-07 at 11 29 27

**Figure 2** UX mismatch: I totally did not expect to see _so many_ URLs to test. This flow I am running is definitely not the flow for which the original issue was written and I'm unsure whether it solves the original problem.

PR #483, opened 2022-04-21
Add support for sharing a link to the OONI Probe app 

aanorbel:
Fixes https://github.com/ooni/probe/issues/1997

## Proposed Changes

  - Add Support for [Share Extensions](https://developer.apple.com/design/human-interface-guidelines/ios/extensions/sharing-and-actions/) to the app.
  - Update existing methods related to running tests in `OoniRunViewController`

| .      | . |
| ----------- | ----------- |
| Simulator Screen Shot - iPhone 13 Pro Max - 2022-04-22 at 10 13 50  |  Simulator Screen Shot - iPhone 13 Pro Max - 2022-04-22 at 10 13 55       |


ooniprobe/View/OONIRun/OoniRunViewController.m
```diff
@@ -49,27 +50,35 @@ -(void)handleUrlScheme{
      NSLog(@"dict: %@", dict);
     
     NSString *action;
-    if ([[url host] isEqualToString:@"run.ooni.io"])
+    NSUserDefaults *usrInfo = [[NSUserDefaults alloc] initWithSuiteName:@"group.org.openobservatory.ooniprobe"];
+
+    if ([[url host] isEqualToString:@"run.ooni.io"]) {
         action = [[url path] substringFromIndex:1];
```
bassosimone, line 56: I don't think this diff is correct. Sadly, the original code lacks documentation about what it is doing here.

Though, AFAICT, the first case in the `if` (i.e., the one checking for the host being `run.ooni.io`) is the case of links obtained from `run.ooni.io`. The third case (i.e., the one where we check whether the host is equal to `nettest` _could_ be an ancient format used by OONI Run, but I am not sure).

What I find wrong is that, before the change you introduced, there was a check for the minimum version performed over the parameters. Now, this check _only_ applies to the third case (host equal to `nettest`) and not to the first case (host equal to `run.ooni.io`), which however is _the common case_ for modern OONI Run links.

I would suggest to keep the original algorithm and adding the case of `usrInfo` contains `incomingURL` perhaps before and entirely as a separate code path.

I would also suggest to try to document a bit better what we are doing here. My impression is that we do not fully understand what this class is doing, so it seems a good time to figure out where is the OONI Run documentation and ensure that this code references the documentation.

ooniprobe/View/OONIRun/OoniRunViewController.m
```diff
@@ -49,27 +50,35 @@ -(void)handleUrlScheme{
...
-    else
+    } else if ([[[usrInfo dictionaryRepresentation] allKeys] containsObject:@"incomingURL"] && dict[@"tn"] == nil) {
```
bassosimone, line 57: What is the purpose of checking for `dict[@"th"] == nil`?

share/ShareViewController.swift
```diff
@@ -0,0 +1,126 @@
...
+        self.extensionContext!.completeRequest(returningItems: [], completionHandler: nil)
+    }
+
+    override func configurationItems() -> [Any]! {
+        // To add configuration options via table cells at the bottom of the sheet, return an array of SLComposeSheetConfigurationItem here.
+        return []
+    }
+
+}
+*/
```
bassosimone, line 34:
```suggestion
```

Please, do not include large chunks of uncommitted code in merge requests.

We don't want to litter the main branch with uncommitted code, unless there's a good reason for doing that.

share/ShareViewController.swift
```diff
@@ -0,0 +1,126 @@
...
+
+
+class ShareViewController: UIViewController {
```
bassosimone, line 37: Please, add documentation for this class explaining how it interacts with the rest of the app.

bassosimone (changes requested):
I am not convinced the refactored algorithm is correct, so I think we should discuss that.

In my review, I have also highlighted other improvements for this diff.

ooniprobe/View/OONIRun/OoniRunViewController.m
aanorbel: Once the URL is shared, the next OONI Run launch still contains `incomingURL` in `usrInfo`.
Which means the only way to be sure its share action is also to check that `dict[@"tn"]` which is available only when the app is launched through OONI Run Link is absent.

ooniprobe/View/OONIRun/OoniRunViewController.m
aanorbel: When OONI Run is launched from the browser, the actual URL which reaches the app is `ooni://nettest?tn=web_connectivity&ta=...`. On processing, the actual hostname is `nettest`.

bassosimone (changes requested):
I am still not happy about the algorithm we have. I think it's pretty fragile and I am also wondering whether we should have two distinct functions for handling two distinct cases (i.e., sharing of a OONI Run link and sharing a URL with the app for the purpose of testing the URL itself). I would argue this change could make the code cleaner because it would not confuse the reader about what is happening and it would not conflate two pretty distinct flows into the same algorithm.

TBH, I also have pretty serious reservations about how this whole class was originally implemented. In particular, it blows my mind quite a bit that we're modifying the state of the `OONIRunViewController` and I would be much happier if we created a new instance every time, with a pristine state, and set the state and run the test using such a state.

With the current design, I am worried about the fact that we are modifying the existing state every time we are called. Now, it may be that we're safe because the life cycle of the view controller is such that a new view controller is created every time it's needed. But I still find it to be bad form to have mutable state when it's not needed. (If I am wrong and we actually cannot write this algorithm without mutating state, please educate me!).

The minimum set of changes that I think we need to apply to make this algorithm reasonable is the following:

```diff
--- A	2022-06-27 12:50:38.000000000 +0200
+++ B	2022-06-27 12:56:09.000000000 +0200
@@ -1,12 +1,21 @@
 -(void)handleUrlScheme{
     NSDictionary *dict = [DictionaryUtility parseQueryString:[url query]];
      //Logging
-     NSLog(@"url recieved: %@", url);
+     NSLog(@"url received: %@", url);
      NSLog(@"query string: %@", [url query]);
      NSLog(@"host: %@", [url host]);
      NSLog(@"url path: %@", [url path]);
      NSLog(@"dict: %@", dict);
     
+    // Case #1: OONI run link
+    //
+    // - https://run.ooni.io/nettest?... is the URL you receive when
+    // the app does not correctly handle deeplinks
+    //
+    // - ooni://nettest?... is the URL you receive when the app
+    // correctly handles deeplinks
+    //
+    // Empirically, the `https://run.ooni.io/...` case could happen.
     NSString *action;
     if ([[url host] isEqualToString:@"run.ooni.io"]) {
         action = [[url path] substringFromIndex:1];
@@ -23,12 +32,16 @@
                 [self setTestName:[parameters objectForKey:@"tn"]];
                 if ([parameters objectForKey:@"ta"])
                     [self setTestArguments:[parameters objectForKey:@"ta"]];
+                [self showTestScreen];
             } else {
                 [self showErrorScreen];
             }
+            return;
         }
     }
 
+    // case #2: the user shares a link with the OONI app, so we are
+    // going to test the user-provided URL using webconnectivity
     NSUserDefaults *usrInfo = [[NSUserDefaults alloc] initWithSuiteName:@"group.org.openobservatory.ooniprobe"];
     if ([[[usrInfo dictionaryRepresentation] allKeys] containsObject:@"incomingURL"] && dict[@"tn"] == nil) {
         action = @"nettest";
@@ -37,11 +50,7 @@
         [self setTestArguments:@{
                 @"urls": @[[usrInfo valueForKey:@"incomingURL"]]
         }];
-    }
-
-    if ([action isEqualToString:@"nettest"]) {
         [self showTestScreen];
-    } else {
-        [self showErrorScreen];
+	return;
     }
 }
```

I propose this set of changes because it gives me some creeps to think that we are falling through from one case to the other and I think this code is written in a way that will only make it worst to maintain over time.

To summarize:

1. I would recommend to redesign this code to (a) have two distinct entry points for the two distinct user stories we're supporting and (b) avoid mutating itself but rather create some new state and explicitly use it _only_ for a single invocation;

2. if that is not possible or deemed to much work, which may be the case given that we have many issues on our plate, then I would at least modify the diff as proposed to start decoupling a little bit the two distinct cases (but this is not my preferred option as I think we would benefit from more clean separation here);

3. if we choose to keep the current design, I would like to see an issue explaining the shortcomings of this code and proposing a more clear design.

I am sorry we are continuing to propose changes to this pull request and we are not yet ready to green light it. Unfortunately, the more I look into the original code that we are modifying in this pull request, the less I am convinced about it being good code.

bassosimone (commented):
Looks a significant improvement with what we had before! 🥳 

That's fantastic!

Yet, to complete my review, I think I need a little help to read and understand the code.

share/ShareViewController.swift
```diff
@@ -0,0 +1,106 @@
...
+        } else if itemProvider.hasItemConformingToTypeIdentifier(typeURL) {
+            handleIncomingURL(itemProvider: itemProvider)
+        } else {
+            print("Error: No url or text found")
+            self.extensionContext?.completeRequest(returningItems: nil, completionHandler: nil)
+        }
+    }
+
+    private func handleIncomingText(itemProvider: NSItemProvider) {
+        itemProvider.loadItem(forTypeIdentifier: typeText, options: nil) { (item, error) in
```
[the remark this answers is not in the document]
aanorbel: yeah thats correct

share/ShareViewController.swift
```diff
@@ -0,0 +1,106 @@
...
+
+            let json = NSString(data: data, encoding: String.Encoding.utf8.rawValue)
+            let queryItems = [
+                URLQueryItem(name: "mv", value: "1.2.0"),
+                URLQueryItem(name: "tn", value: "web_connectivity"),
+                URLQueryItem(name: "ta", value: json as String?)
+            ]
+            var urlComps = URLComponents(string: appURL)!
+            urlComps.queryItems = queryItems
+            let result = urlComps.url!
```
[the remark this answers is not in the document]
aanorbel: yeah correct

PR #490, opened 2022-06-13
Unable to view data for uploaded measurements.

aanorbel:
Fixes  https://github.com/ooni/probe/issues/2140

## Proposed Changes

  - Change API call to fetch data from the backend.

IMG_4EFACBCC9549-1


bassosimone (approved):
🐳

PR #491, opened 2022-06-16
Wrong ASN displayed in successful results of a re-run failed test when different network is selected.

aanorbel:
Fixes  https://github.com/ooni/probe/issues/851

## Proposed Changes

  -  Update `Measurement`  table to add  `rerun_network` column.
  - Annotate  `AbstractTest` before rerun is triggered.
  - Capture network during `status.geoip_lookup` event and save to class instance.
  - Update measurement to include new network during `measurement` probe-enigine evennt


bassosimone (approved):
🐳

PR #493, opened 2022-06-30
Feat: append `-unattended` to software_name used for unattended runs

aanorbel:
Fixes https://github.com/ooni/probe/issues/2167

## Proposed Changes

  - Update `getTestList` to set `autoRun` of tests to `YES` when appropriate.
  - Update `AbstractTest` to set `settings.annotations[@"origin"]` to `autorun` and also append `-unattended` to `settings.options.software_name`

### DIFF
--- a/ooniprobe/Test/Suite/AbstractSuite.m
+++ b/ooniprobe/Test/Suite/AbstractSuite.m
@@ -100,7 +100,7 @@
 }
 
 -(NSArray*)getTestList {
-    if (self.autoRun){
+    if (self.autoRun) {
         for (int i = 0; i < [self.testList count]; ++i) {
             AbstractTest *current = self.testList[(NSUInteger) i];
             current.autoRun = YES;
--- a/ooniprobe/Test/Suite/AbstractSuite.m
+++ b/ooniprobe/Test/Suite/AbstractSuite.m
@@ -100,6 +100,13 @@
 }
 
 -(NSArray*)getTestList {
+    if (self.autoRUn){
+        for (int i = 0; i < [self.testList count]; ++i) {
+            AbstractTest *current = self.testList[(NSUInteger) i];
+            current.autoRUn = YES;
+            self.testList[(NSUInteger) i] = current;
+        }
+    }
     return self.testList;
 }
 
--- a/ooniprobe/Test/Suite/AbstractSuite.m
+++ b/ooniprobe/Test/Suite/AbstractSuite.m
@@ -100,9 +100,11 @@
 }
 
 -(NSArray*)getTestList {
-    if (self.autoRUn){
+    if (self.autoRun){
         for (int i = 0; i < [self.testList count]; ++i) {
-            self.testList[(NSUInteger) i].autoRUn = YES;
+            AbstractTest *current = self.testList[(NSUInteger) i];
+            current.autoRun = YES;
+            self.testList[(NSUInteger) i] = current;
         }
     }
     return self.testList;
--- a/ooniprobe/Test/Suite/AbstractSuite.m
+++ b/ooniprobe/Test/Suite/AbstractSuite.m
@@ -102,9 +102,7 @@
 -(NSArray*)getTestList {
     if (self.autoRUn){
         for (int i = 0; i < [self.testList count]; ++i) {
-            AbstractTest *current = self.testList[(NSUInteger) i];
-            current.autoRUn = YES;
-            self.testList[(NSUInteger) i] = current;
+            self.testList[(NSUInteger) i].autoRUn = YES;
         }
     }
     return self.testList;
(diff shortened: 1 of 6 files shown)


ooniprobe/Test/Suite/AbstractSuite.m
```diff
@@ -100,6 +100,13 @@ - (void)showLocalNotification {
 }
 
 -(NSArray*)getTestList {
+    if (self.autoRUn){
+        for (int i = 0; i < [self.testList count]; ++i) {
+            AbstractTest *current = self.testList[(NSUInteger) i];
+            current.autoRUn = YES;
+            self.testList[(NSUInteger) i] = current;
```
bassosimone, line 107:
```suggestion
            self.testList[(NSUInteger) i].autoRUn = YES;
```

A bunch of comments and questions:

1. beware of the capitalization: `RUn` should be `Run`
2. would the code above work?
3. what is the lifecycle of this `self.testList`?
4. what is the implication of setting `autoRun` to `YES`? could it stick to future runs?
5. is there a way to write this code to avoid mutating the content of `self`?

ooniprobe/Test/Test/AbstractTest.m
```diff
@@ -57,6 +57,11 @@ - (Measurement*)createMeasurementObject{
 
 -(void)prepareRun{
     self.settings = [Settings new];
+    if (self.autoRUn){
```
bassosimone, line 60:
```suggestion
    if (self.autoRUn) {
```

ooniprobe/Test/Test/AbstractTest.m
```diff
@@ -57,6 +57,11 @@ - (Measurement*)createMeasurementObject{
...
+        self.settings.annotations[@"origin"] = @"autorun";
+        self.settings.options.software_name = [NSString stringWithFormat:@"%@%@",SOFTWARE_NAME,@"-unattended"];
+        NSLog(self.settings.annotations[@"origin"]);
```
bassosimone, line 63:
```suggestion
```

I think we don't want to commit this logging message

ooniprobe/Utility/BackgroundTask.m
```diff
@@ -77,7 +77,9 @@ + (void)checkIn {
     [testSuite setStoreDB:NO];
     AbstractTest *test = [[AbstractTest alloc] initTest:testName];
     [test setStoreDB:NO];
+    [test setAutoRUn:YES];
     [test setAnnotation:YES];
+    test.settings.annotations[@"origin"] = @"autorun";
```
bassosimone, line 82: I see you're already doing this in `AbstractTest.m`. Do we need to do it twice or can we avoid doing it here?

bassosimone (commented):
I have a bunch of questions about the code that would be instrumental to give me a better understanding of the code we're changing and overall the changes we're making

ooniprobe/Test/Suite/AbstractSuite.m
aanorbel: this would not work. the direct assignment posses some problems with the types. i tried that.

ooniprobe/Test/Suite/AbstractSuite.m
```diff
@@ -100,6 +100,13 @@ - (void)showLocalNotification {
 }
 
 -(NSArray*)getTestList {
+    if (self.autoRun){
```
bassosimone, line 103:
```suggestion
    if (self.autoRun) {
```

bassosimone (commented):
I am still a bit confused of the amount of self mutation going on here.

If you don't mind, I'd rather have a chat with you about this topic in our Slack channel.

I'll make sure I come back here and write a summary of our chat afterwards.

ooniprobe/Test/Suite/AbstractSuite.m
```diff
@@ -100,6 +100,13 @@ - (void)showLocalNotification {
 }
 
 -(NSArray*)getTestList {
+    if (self.autoRun) {
+        for (int i = 0; i < [self.testList count]; ++i) {
+            AbstractTest *current = self.testList[(NSUInteger) i];
+            current.autoRun = YES;
+            self.testList[(NSUInteger) i] = current;
+        }
+    }
```
bassosimone, line 109: The current implementation needs to use state to pass information from the suite-specific `getTestList` to the abstract-suite's `getTestList`. It is unclean to use the state to pass information around like this, and we should see to refactor the code in a way that does not require us to use state here. Ideally, there should be a method that calls the sub-class or whatever `getTestList` and then code that filters and returns this list without using the state to pass information from the more specific class to the more generic class.

bassosimone (approved):
I am going to approve this pull request and comment on what I think it is doing that is not clean and why I think we should refactor the current code in the future.

PR #496, opened 2022-07-15
No warning displayed when starting OONI Run link based test while under a VPN

aanorbel:
Fixes  https://github.com/ooni/probe/issues/2138

## Proposed Changes

  - Update `OoniRunViewController` to display as `UIModalPresentationFullScreen`.
  - Check VPN availability before `prepareForSegue`


bassosimone (approved):
🐳

PR #507, opened 2023-01-19
Background Processing update

aanorbel:
Fixes  https://github.com/ooni/probe/issues/2175

## Proposed Changes

  - Call `interruptTest` in expiration handler


bassosimone (approved):
🐳

PR #513, opened 2023-03-20
Fix: Correct null url introduced after change from `downloadUrls` to `checkIn` in #510

aanorbel:
Fix: Correct null url introduced after change from `downloadUrls` to `checkIn` in #510

## Proposed Changes

  - Add code to save URLs in `OONIApi.checkInCallback` after `OONIApi.checkIn` is called.


bassosimone (approved):
🐳

PR #515, opened 2023-05-09
Fix: background task notifications are sometimes undefined and lead to app crash

aanorbel:
Fixes  https://github.com/ooni/probe/issues/2349

## Proposed Changes

  - Add default value for `getNameForTest` when `result` is not saved. 

|.|.|
|-|-|
| image | image |
| image | .


bassosimone (approved):
🐳

PR #517, opened 2023-05-29
fix: show remaining time as h-m-s instead of seconds

aanorbel:
Fixes  https://github.com/ooni/probe/issues/2454

## Proposed Changes

  - Converted time remaining to human-readable format.

### DIFF
--- a/ooniprobe/View/RunTest/TestRunningViewController.m
+++ b/ooniprobe/View/RunTest/TestRunningViewController.m
@@ -78,8 +78,7 @@
         //We don't want to show -1 seconds before downloading the URL list
         if (self.totalRuntime <= [MAX_RUNTIME_DISABLED intValue])
             return;
-        NSString *time = NSLocalizedFormatString(@"Dashboard.Running.Seconds", [NSString stringWithFormat:@"%d", self.totalRuntime]);
-        [self.timeLabel setText:time];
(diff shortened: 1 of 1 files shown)


ooniprobe/View/RunTest/TestRunningViewController.m
```diff
@@ -78,8 +78,7 @@ -(void)setRuntime{
         //We don't want to show -1 seconds before downloading the URL list
         if (self.totalRuntime <= [MAX_RUNTIME_DISABLED intValue])
             return;
-        NSString *time = NSLocalizedFormatString(@"Dashboard.Running.Seconds", [NSString stringWithFormat:@"%d", self.totalRuntime]);
-        [self.timeLabel setText:time];
+        [self.timeLabel setText:[NSString stringWithFormat:@"%dm %ds", self.totalRuntime/60, self.totalRuntime%60]];
```
bassosimone, line 81: Below you're using `%ld` and here `%d`. I want to make sure this is intentional.

aanorbel, line 81:
```suggestion
        [self.timeLabel setText:[NSString stringWithFormat:@"%ldm %lds", self.totalRuntime/60, self.totalRuntime%60]];
```

PR #523, opened 2023-06-19
fix: bug report email refers to a wrong iPhone model 

aanorbel:
Fixes  https://github.com/ooni/probe/issues/2486

## Proposed Changes

  - Updated bug report email to use model name instead of model identifier.


bassosimone (changes requested):
I would like to discourage us from merging this issue. By creating this list of mappings between the model identifier and the model name, it means we're going to need to keep this list up to date. People with new models will further be confused.

If the model identifier is sufficient to us to understand what device the code is actually using, which seems the case given the above mapping, I think it's fine to keep the code as it was before. I understand how seeing the wrong "model" could create confusion to users reporting bugs, but I also reckon we have little bandwidth to track all the issues we already have, so I would really like to avoid having the N+1 issue we need to track and make sure we don't have an N+1 piece of code we need to remember to keep up to date.

The only conditions under which I think it's acceptable to merge this patch are the following:

1. we update the [release QA procedure](https://github.com/ooni/probe/blob/master/.github/ISSUE_TEMPLATE/mobile-qa.md) to remind us to update this list;
2. we automatically fetch and update this list from some authoritative place.

Thank you!

PR #524, opened 2023-07-18
chore: use probe-cli@v3.18.1

bassosimone:
Part of https://github.com/ooni/probe/issues/2503


bassosimone:
I am trying to figure out why the CI checks fail here.

bassosimone:
I am starting to understand the issue. These symbols are missing:

```
 Testing failed:
	Undefined symbol: _res_9_nclose
	Undefined symbol: _res_9_ninit
	Undefined symbol: _res_9_nsearch
	Testing cancelled because the build failed.
```

I am wondering why. We incremented the Go version we're using so this may be the reason(?).

PR #527, opened 2023-07-25
fix: blank suite names in results filtering

aanorbel:
Fixes  https://github.com/ooni/probe/issues/2510

## Proposed Changes

  -  Updates `TestUtility#getTestTypes`  mutate returned array to have only the necessary items


bassosimone (changes requested):
I am not sure hardcoding the list is a long term fix for this issue. We're about to introduce OONI Run v2 descriptors, which will probably also add to this list. Perhaps, a better fix would be to keep the previous code and make sure we drop empty entries from the list? More in detail, I think we should get the keys and then return to the caller a new NSArray populated only with non-empty fields.

hellais (approved):
LGTM 🐳

PR #529, opened 2023-07-28
Fix: invalid handling of malformed URL

aanorbel:
Fixes  https://github.com/ooni/probe/issues/2516

## Proposed Changes

  - Updated app to show invalid arguments if test arguments cannot be parsed.


hellais (approved):
LGTM 🐳

PR #542, opened 2023-11-29
fix: test results show ongoing test instead of older measurements while test is running

aanorbel:
Fixes https://github.com/ooni/probe/issues/2638

## Proposed Changes

  - Check if `result.id` from the incoming update matches `result.id` of the currently visible test result.


bassosimone (approved):
🐳

PR #552, opened 2024-01-30
fix: The name of the page disappears from the header on iOS

aanorbel:
Fixes  https://github.com/ooni/probe/issues/2648

## Proposed Changes

  - Updates `SettingsTableViewController` to set title on `viewWillAppear`.


bassosimone (approved):
🐳

PR #561, opened 2024-06-03
fix: `UploadFooterViewController` cause app to crash when upload fails with error

aanorbel:
Fixes  https://github.com/ooni/probe/issues/2710

## Proposed Changes

  -  Replace `[error dictionary]` with empty dictionary (`[error dictionary]` seems to fails consistently.)


bassosimone (approved):
🐳

PR #562, opened 2024-06-06
fix: Websites categories action causes crash.

aanorbel:
Fixes  https://github.com/ooni/probe/issues/2744

## Proposed Changes

  - Add extra controls for alert on iPad.


bassosimone (approved):
🐳

